PR #16, opened 2020-10-02
big changes

cscherrer:
Lots of changes here:

- Added `sampletype`, as suggested by @phipsgabler (`eltype` doesn't cut it)
- Removed the `X` type parameter in a few places (`AbstractMeasure`, `TrivialMeasure`, parameterized measures)
- Started a `basemeasures.jl` to define what a `baseMeasure` is and import things
- Added `representative`, i.e., of the equivalence class wrt absolute continuity. In general this is different than a base measure
- Added `isprimitive`, which just indicates a measure is not defined in terms of others
- Removed `×` for measures, as @sethaxen pointed out it clashes with `LinearAlgebra.×`, and @mschauer that \otimes is more standard
- Made `Density <: Function`, because why not
- Added support for `PushForward` in terms of `Bijectors`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 1.3
+  - 1.5
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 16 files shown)


cscherrer:
I'm going to merge this. There will be plenty more to fix, but we're still in the early stages anyway, and it's at least a net gain :)

PR #20, opened 2020-10-08
Add (measure-) kernel

mschauer:
I made some fixes to Dirac along the way.


src/probability/dirac.jl
```diff
@@ -2,21 +2,17 @@
 export Dirac
 
 struct Dirac{X} <: AbstractMeasure
-    supp :: X
```
mschauer, line 5: Rename to `x` as `supp` seems to be a set, not a value.

src/probability/dirac.jl
```diff
@@ -2,21 +2,17 @@
...
+sampletype(μ::Dirac{X}) where {X} = X
 
 function (μ::Dirac{X})(s) where {X}
-    μ.supp ∈ s && return 1
+    μ.x ∈ s && return 1
     return 0
 end
+isprimitive(::Dirac) = true
 
-function logdensity(μ::Dirac{X}, x::X)
```
mschauer, line 17: not a density, rather a primitive

src/rand.jl
```diff
@@ -0,0 +1 @@
+Base.rand(μ::AbstractMeasure) = rand(Random.GLOBAL_RNG, μ)
```
mschauer, line 1: Perhaps a bit nicer this way

test/runtests.jl
```diff
@@ -4,3 +4,8 @@ using Test
 @testset "MeasureTheory.jl" begin
     # Write your own tests here.
 end
+
+@testset "Kernel" begin
+    κ = MeasureTheory.kernel(x->(x,), MeasureTheory.Dirac)
+    @test rand(κ(1.1)) == 1.1
```
mschauer, line 10: Testing the special case of the deterministic kernel, just to see

src/rand.jl
cscherrer: Absolutely,! Oh, I guess we can just have this in one place, and then for a particular measure just define `rand(::AbstractRNG, μ)`. So we won't have to have quite as much duplication as I had thought. Nice :)

src/probability/dirac.jl
cscherrer: Wouldn't `Dirac(1.0)` make sense? OTOH we may sometimes want this to be represented as `::Dirac{Float64}`, or possibly (for static analysis) as `::Dirac{Val{1.0}}`. But the first seems simpler to start with.

src/probability/dirac.jl
cscherrer: Good catch, thanks! It will have a density wrt a discrete measure over a large support, but the for I use here assumes there's a base measure, which is not right.

test/runtests.jl
cscherrer: Great!!

cscherrer:
Thanks @mschauer ! I added a comment on the `supp` thing - I had been thinking the parameter for a Dirac would be a value. We can discuss that some more if you disagree, everything else is good to go.

src/probability/dirac.jl
cscherrer: Then again, the value determines a singleton set, which _is_ the support. And maybe `supp` would carry over more uniformly to other measures? So, ok maybe I've convinced myself I agree

cscherrer (approved):

PR #21, opened 2020-10-22
Improve a bit the kernel interface

mschauer:


mschauer:
I am now able to define a simple state space system with this (counts as a test too):

https://github.com/mschauer/Mitosis.jl/blob/master/test/statespace.jl

mschauer:
As a next thing, I made the measure a type parameter so I can dispatch on it.

mschauer:
Good to go for now

mschauer:
Bump.

PR #22, opened 2020-10-29
 Improve a bit the kernel interface and Mitosis compatibility 

mschauer:


cscherrer (commented):
Hi @mschauer , thanks for the PR

Just so I understand the changes here... Originally we had 
```julia
kernel(f,m)(x) == m(f(x))
```

With the PR, I see two big changes:
1. `kernel(f,m)` has become `kernel(m,f)`
2. `m` is no longer a measure, but is instead a type constructor for a measure

Is that right?

It's convenient to have a function as the first argument when possible, so we get `do` notation, which can be pretty handy. That's tricky if we need multiple functions, but I'm not sure we really need that in this case.

Instead of a tuple `ops` of functions, would it be flexible enough to have a single function that returns the needed parameterization? Something like

```julia
struct Kernel{F,M}
    f::F
    m::M
end

(k::kernel)(x) = k.m(k.f(x))
```
?

mschauer:
I still think it makes sense to move the measure into a type parameter but I think I have an idea for the other question

mschauer:
How do you like
```
julia> using Distributions

julia> (kernel(Normal) do x; (x, 2x) end)(1.0)
Normal{Float64}(μ=1.0, σ=2.0)

julia> kernel((x->x, x->2x), Normal)(1.0)
Normal{Float64}(μ=1.0, σ=2.0)
```

cscherrer:
> I still think it makes sense to move the measure into a type parameter

I agree this is nice. It just seems to me we might sometimes need `f::Function`. Anyway, if we do it seems like an easy enough extension.

Using the `MeasureTheory` interface, would it also work as

```julia
julia> kernel(x -> (μ=x, σ=2x), Normal)(1.0)
Normal{Float64}(μ=1.0, σ=2.0)
```
and
```julia
julia> kernel((μ = x -> x, σ = x -> 2x), Normal)(1.0)
Normal{Float64}(μ=1.0, σ=2.0)
```
?

mschauer:
>  would it also work as

Yes, I wasn't sure if for this case it makes sense to retain the old form
```
 kernel(Normal, μ = x -> x, σ = x -> 2x)(1.0)
```

mschauer:
Ok, I put the function argument first now and left the convenient call signature with the named arguments in.

PR #44, opened 2021-01-13
@μσ_methods

cscherrer:
This PR adds a macro `@μσ_methods` that makes it easier to add methods for location-scale families. For example, we can start with just `StudentT(ν)`, and then (editing out some line-number stuff)
```julia
julia> @macroexpand MeasureTheory.@μσ_methods StudentT(ν)
quote
    function rand(rng::AbstractRNG, d::StudentT{(:ν, :μ, :σ)}, x)
        d.σ * rand(rng, StudentT(d.ν)) + d.μ
    end

    function logdensity(d::StudentT{(:ν, :μ, :σ)}, x)
        z = (x - d.μ) / d.σ
        return logdensity(StudentT(d.ν), z) - log(d.σ)
    end

    function rand(rng::AbstractRNG, d::StudentT{(:ν, :σ)}, x)
        d.σ * rand(rng, StudentT(d.ν))
    end

    function logdensity(d::StudentT{(:ν, :σ)}, x)
        z = x / d.σ
        return logdensity(StudentT(d.ν), z) - log(d.σ)
    end

    function rand(rng::AbstractRNG, d::StudentT{(:ν, :μ)}, x)
        rand(rng, StudentT(d.ν)) + d.μ
    end

    function logdensity(d::StudentT{(:ν, :μ)}, x)
        z = x - d.μ
        return logdensity(StudentT(d.ν), z)
    end
end
```

`rand` methods aren't quite working right yet, but it's pretty close.


mschauer:
This is neat, actually.

PR #73, opened 2021-03-07
Disambiguate Product constructor

mschauer:
If the argument to the constructor `ProductMeasure(data...)` has only one element the meaning of the argument changes. This prevents from starting accumulation of a product measure with an empty product
```
acc = ProductMeasure(())
...
white ...
    acc = acc*ν
end
```


cscherrer (approved):
Looks good, thanks @mschauer !

PR #84, opened 2021-03-28
Markov Chains

cscherrer:
This PR adds two new structs:

- `Chain` represents a Markov Chain
- `RandChain` represents an instance of calling `rand` on a `Chain`

Things aren't quite right yet - we'll need to store the random seed as well to make sure the random choices are treated as being made when `rand` is called. For example, this is not great:

```
julia> mc = Chain(Normal(μ=0.0)) do x Normal(μ=x) end
Chain{}(var"#7#8"(), Normal(μ = 0.0,))

julia> r = rand(mc)
RandChain{}(Random._GLOBAL_RNG(), Chain{}(var"#7#8"(), Normal(μ = 0.0,)))

julia> take(r, 10) |> collect
10-element Vector{Any}:
  0.32269884167170293
  0.08670535551081499
 -1.0116170752955713
 -2.035667144109519
 -2.376649536399607
 -3.1358050779220843
 -1.8979886304591245
 -1.8469492650256318
 -3.7145560325153397
 -3.7000372946869535

julia> take(r, 10) |> collect
10-element Vector{Any}:
  0.13998784174918052
  0.21616974318137017
  0.05482135811998662
 -0.21706907568347017
  0.39383245114988713
  1.6763896719921025
  1.4151968406002067
  0.7521407210667452
  1.3037282896135198
  0.551151422401246
```


cscherrer:
@mschauer does this use of DynamicIterators look reasonable? Any changes you'd suggest?

cscherrer:
In this model:
```julia

hmm = @model begin
    ε ~ Exponential() #  transition
    σ ~ Exponential() # Observation noise
    x ~ Chain(Normal()) do xj
        Normal(xj, ε)
    end

    y ~ For(x) do xj
        Normal(xj, σ)
    end
end
```

We currently get
```julia
julia> rand(hmm())
ERROR: MethodError: no method matching For(::GeneralizedGenerated.Closure{function = (σ, xj;) -> begin
    (Main).Normal(xj, σ)
end, Tuple{Float64}}, ::RandChain{})
```

I guess to make this general we could add
```
For(f, ::DynamicIterator)
```

src/combinators/chain.jl
```diff
@@ -0,0 +1,107 @@
+using ConcreteStructs
+using DynamicIterators
+using DynamicIterators: dub
+import Base
```
mschauer, line 4: What does this do?

cscherrer: `ConcreteStructs` is pretty great:
https://github.com/jonniedie/ConcreteStructs.jl

Some of this is vestigial, I was using `dub` for a while before I realized an `Evolution` doesn't need it

cscherrer:
@mschauer I still have `RandChain` here. Let's keep this until we have implemented your DyanmicIterator suggestion. Then we can compare them and drop `RandChain` if there's no need for it

cscherrer:
I got this behavior working, which I think it important:
```julia
julia> using Base.Iterators

julia> using MeasureTheory
[ Info: Precompiling MeasureTheory [eadaa1a4-d27c-401d-8699-e962e1bbc33b]

julia> mc = Chain(Normal(μ=0.0)) do x Normal(μ=x) end
Chain{}(var"#5#6"(), Normal(μ = 0.0,))

julia> r = rand(mc)
RandChain{}(MersenneTwister(0xc197cfa1f880e945bf16778a688fa47f, (0, 2256, 0, 338, 1002, 1)), 0x6a9ea19d77a18c5a, Chain{}(var"#5#6"(), Normal(μ = 0.0,)))

julia> @show collect(take(r,10)) == collect(take(r,10))
collect(take(r, 10)) == collect(take(r, 10)) = true
```

cscherrer:
Also
```julia
julia> r2 = rand(mc)
RandChain{}(MersenneTwister(0xc197cfa1f880e945bf16778a688fa47f, (0, 2256, 0, 338, 1002, 2)), 0x46b6143dc1ab393b, Chain{}(var"#5#6"(), Normal(μ = 0.0,)))

julia> collect(take(r,10)) - collect(take(r2,10))
10-element Vector{Float64}:
 -0.6187317698864626
 -0.20297245943720754
  2.380162589168478
  2.9262425431469956
  2.4053912122729195
  1.1381388252174576
  0.5684716250935393
  2.5316455291146394
  2.6746813220844023
  1.2445568557937412
```

test/runtests.jl
```diff
@@ -86,3 +90,66 @@ end
...
     @test logdensity(Dirac(0.3), 0.4) == -Inf
 end
+
+import MeasureTheory.:⋅
+function ⋅(μ::Normal, kernel) 
+    m = kernel(μ)
+    Normal(μ = m.μ.μ, σ = sqrt(m.μ.σ^2 + m.σ^2))
+end
+
+"""
```
mschauer, line 100: Via https://github.com/mschauer/Mitosis.jl/blob/master/src/markov.jl

cscherrer:
Looks like we need an `evolve`:
```julia
julia> r = rand(mc)
MeasureTheory.Realized{UInt64, MersenneTwister, Chain{}}(0x84ea5f216e763857, MersenneTwister(0xdbf61e8e9fdb390194d55a13ce3542d0, (0, 2256, 0, 311, 1002, 6)), Chain{}(var"#5#6"(), Normal(μ = 0.0,)))

julia> d2 = For(r) do x Normal(μ=x) end
DynamicFor{}(var"#13#14"(), MeasureTheory.Realized{UInt64, MersenneTwister, Chain{}}(0x84ea5f216e763857, MersenneTwister(0xdbf61e8e9fdb390194d55a13ce3542d0, (0, 2256, 0, 311, 1002, 6)), Chain{}(var"#5#6"(), Normal(μ = 0.0,))))

julia> rand(d2)
MeasureTheory.Realized{UInt64, MersenneTwister, DynamicFor{}}(0x7e27b45779feda24, MersenneTwister(0xdbf61e8e9fdb390194d55a13ce3542d0, (0, 2256, 0, 311, 1002, 7)), DynamicFor{}(var"#13#14"(), MeasureTheory.Realized{UInt64, MersenneTwister, Chain{}}(0x84ea5f216e763857, MersenneTwister(0xdbf61e8e9fdb390194d55a13ce3542d0, (0, 2256, 0, 311, 1002, 6)), Chain{}(var"#5#6"(), Normal(μ = 0.0,)))))

julia> take(rand(d2), 10) |> collect
ERROR: MethodError: no method matching evolve(::DynamicFor{})
```

cscherrer:
Hmm, but I guess it's only an `evolve` for a Markov chain, and that's no longer the case. I guess we need something similar to a `Base.Generator` that also holds the random state

mschauer:
What’s the stack trace? Looks like a missing method or a to wide method signature

mschauer:
This is tricky, currently Realized works for evolutions where the value is the state, but I hope I know the right incantation

cscherrer:
```julia
julia> take(rand(d2), 10) |> collect
ERROR: MethodError: no method matching evolve(::DynamicFor{})
Closest candidates are:
  evolve(::StepRange, ::Nothing) at /home/chad/.julia/packages/DynamicIterators/oF3Tg/src/evolution.jl:28
  evolve(::StepRange, ::Any) at /home/chad/.julia/packages/DynamicIterators/oF3Tg/src/evolution.jl:24
  evolve(::DynamicIterators.Synchronize, ::T) where T at /home/chad/.julia/packages/DynamicIterators/oF3Tg/src/combinators.jl:153
  ...
Stacktrace:
 [1] dyniterate(rv::MeasureTheory.Realized{UInt64, MersenneTwister, DynamicFor{}}, #unused#::Nothing)
   @ MeasureTheory ~/git/MeasureTheory.jl/src/combinators/chain.jl:52
 [2] iterate(E::MeasureTheory.Realized{UInt64, MersenneTwister, DynamicFor{}})
   @ MeasureTheory ~/git/MeasureTheory.jl/src/combinators/chain.jl:47
 [3] iterate
   @ ./iterators.jl:673 [inlined]
 [4] iterate
   @ ./iterators.jl:671 [inlined]
 [5] copyto!
   @ ./abstractarray.jl:837 [inlined]
 [6] _collect
   @ ./array.jl:608 [inlined]
 [7] collect
   @ ./array.jl:602 [inlined]
 [8] |>(x::Base.Iterators.Take{MeasureTheory.Realized{UInt64, MersenneTwister, DynamicFor{}}}, f::typeof(collect))
   @ Base ./operators.jl:859
 [9] top-level scope
   @ REPL[22]:1
```

src/combinators/chain.jl
```diff
@@ -0,0 +1,152 @@
...
+For(f, r::Realized) = DynamicFor(f,r)
+
+function Base.rand(rng::AbstractRNG, dfor::DynamicFor)
+    seed = rand(rng, UInt)
+    return Realized(seed, copy(rng), dfor)
+end
+
+function dyniterate(df::DynamicFor, st, args...)
+    (val, state) = dyniterate(df.iter, st, args...)
+    return (f(val), state)
```
mschauer, line 84:
```suggestion
    return (df.κ(val), state)
```

PR #86, opened 2021-04-04
Markov Chains 2 (WIP)

mschauer:


test/runtests.jl
```diff
@@ -86,3 +90,67 @@ end
     @test logdensity(Dirac(0.3), 0.3) == 0.0
     @test logdensity(Dirac(0.3), 0.4) == -Inf
 end
+
+import MeasureTheory.:⋅
+function ⋅(μ::Normal, kernel)
```
cscherrer, line 95: We should have this in the package, not just tests. But we'll need to narrow down the argument types a little bit for it to work

test/runtests.jl
```diff
@@ -86,3 +90,67 @@ end
...
+
+@testset "DynamicFor" begin
+    mc = Chain(Normal(μ=0.0)) do x Normal(μ=x) end
+    @test_broken rand(mc)
+    r = rand(Random.GLOBAL_RNG, mc)
+   
+    # Check that `r` is now deterministic
+    @test logdensity(mc, take(r, 100)) == logdensity(mc, take(r, 100))
+    
+    d2 = For(r) do x Normal(μ=x) end
```
cscherrer, line 126: From this construction, we need to be sure `rand(d2)` works

cscherrer (commented):
Thanks @mschauer , looks like we're getting there!

PR #93, opened 2021-05-28
KeywordCalls compatibility

cscherrer:
# Changes

**Dependencies**
- now using KeywordCalls.jl
- Early stages of trait interface for `IsPrimitive` and `IsRepresentative`, using SimpleTraits.jl (see `traits.jl`)
- New Distributions.jl version (no real changes here from that)
- New TransformVariables has breaking changes, see `transforms.jl`

**Renaming**
- `@measure` is too vague, changed to `@parameterized`
- `ElementwiseProductMeasure` is a bit much, it's now `PointwiseProductMeasure`
- `Likelihood` is now `LogLikelihood`, and it's callable as a function
- `src/probability` is now `src/parameterized`

**New stuff**
- Added `@σ_methods`, mostly for use with measures on postive reals
- Got `asparams` working better
- **Lots** of comments in https://github.com/cscherrer/MeasureTheory.jl/blob/cs-keywordcalls/src/parameterized/normal.jl
- `@primitive` to declare a measure to have the `IsPrimitive` trait (very simple, really just to save end users from having to worry about traits)


cscherrer:
I had to make some adjustments in KeywordCalls. Best to use version 1.8 of that, or (until that's registered) this branch:
https://github.com/cscherrer/KeywordCalls.jl/pull/23

src/distributions.jl
```diff
@@ -27,16 +27,17 @@ function Base.rand(rng::AbstractRNG, μ::DistributionMeasure{F,S,X}) where {F, S
...
 basemeasure(d::Dists.AbstractMvNormal) = Lebesgue(ℝ)^size(d)
 
 function basemeasure(μ::Dists.Distribution{Dists.Univariate,Dists.Continuous})
     return Lebesgue(ℝ)
 end
 
+∫(::typeof(identity), ::Dists.Distribution) = 1.0
 
 logdensity(μ::Dists.Distribution, x) = Dists.logpdf(μ,x)
+Dists.logpdf(d::AbstractMeasure, x) = logdensity(d,x)
```
mschauer, line 39: should this mimic distribution's `pdf`, which is kind of broken? If we would like to represent the actual behaviour
```
function Dists.logpdf(d::AbstractMeasure, x)
    μ = basemeasure(d)
    if μ ≪ Lebesgue(...) # depending on size?
        logdensity(d,Lebesgue, x)
    elseif μ ≪ CountingMeasure(...)
          logdensity(d,CountingMeasure, x)
    else
          error("logpdf undefined, try logdensity")
    end
end
```

src/macros.jl
```diff
@@ -1,10 +1,8 @@
 
 using MLStyle
-using StatsFuns
 using Random: AbstractRNG
-using StatsFuns: logtwo
 
-export @measure
+export @parameterized
```
mschauer, line 5: I like this

src/parameterized/normal.jl
```diff
@@ -0,0 +1,166 @@
...
+#
+distproxy(d::Normal{(:μ, :σ)}) = Dists.Normal(d.μ, d.σ)
+
+###############################################################################
+# Some distributions have a "standard" version that takes no parameters
+@kwstruct Normal()
+
+# Instead of setting default values, the `@kwstruct` call above makes a
+# parameter-free instance available. The log-density for this is very efficient.
+logdensity(d::Normal{()} , x) = - x^2 / 2
```
mschauer, line 82: Shouldn't we define the three argument form and have `logdensity(d, x) = logdensity(d, basemeasure(d), x)`

src/distributions.jl
cscherrer: Ahh, good point. I'd like to avoid the conditionals if possible, since they can make it hard to extend. But this is the idea of `representative`. There are actually some pending design decisions around that concept. This PR review won't be very visible, so maybe I'll start a new thread in Zulip about it in case anyone else is interested.

src/parameterized/normal.jl
cscherrer: I've thought about the tradeoffs here and currently lean toward this approach. But we should discuss this some more to make sure we've thought through it. I'll start a new issue.

cscherrer: Resolved here:
https://github.com/cscherrer/MeasureTheory.jl/issues/99

src/distributions.jl
cscherrer: @mschauer I added an issue on this here:
https://github.com/cscherrer/MeasureTheory.jl/issues/98

That's maybe not entirely clear - giving a clear description of something as I'm figuring it out is kind of tricky. I think what it comes down to is that we can have

```julia
function Dists.logpdf(d::AbstractMeasure, x)
    b = fix(basemeasure, d)
    return logdensity(d, b)
end
```

This traverses to a fixpoint twice (once to find the fixpoint, and again to compute the `logdensity`). I'm sure we can make this more efficient. When we do, we can add a test

```julia
d = ...
x = rand(d)
@test Dists.logpdf(d, x) == logdensity(d, fix(basemeasure, d), x)
```

What do you think?

cscherrer:
@mschauer I'm going to go ahead with this, with the understanding that some of the "representative measure" stuff is still very unstable

Issue #51, opened 2021-01-31
Design of `rand`


devmotion:
Last week I had some discussion in a PR to Distributions about the design of `rand` (https://github.com/JuliaStats/Distributions.jl/pull/1262#issuecomment-764995042), and I think it applies to MeasureTheory as well.

---

*TL;DR:* It is problematic to prescribe the type of samples of a distribution, both for users and internally. However, similar to `logdensity`, the type of the samples follows naturally if (1) the type of the basic samples from the RNG obtained with `rand`, `randn`, or `randexp`, and (2) the parameters of the distribution of interest are known. Thus I propose to not prescribe the type of the samples of the distribution but the type of the basic samples from the RNG with type `T` in `rand(::AbstractRNG, ::Type{T}, ::Distribution)`.

---

Partly copied from https://github.com/JuliaStats/Distributions.jl/pull/1262:

Basically every sampling procedure at some point starts sampling with `rand`, `randexp`, or `randn` in Random from a uniform distribution in `[0, 1)`, the exponential distribution with rate 1 on `R_{>0}`, or the standard normal distribution on `R`, respectively. The samples from the distribution of interest are obtained by transforming these most basic samples (ie. basically the distribution of interest is a push-forward of the distribution of these basic samples), and the type of the random variates follows automatically from the type of these basic samples and the function that transforms them and involves the parameters of the distribution.

My suggestion would be to allow to specify the type of exactly (and only) these base variates, i.e., to change such "basic calls" to `rand(rng, T)`, `randn(rng, T)` etc. where `T` can be specified by the user. The type of the random variates of the distribution of interest would then be determined both by `T`, the parameters of the distribution, and the transformations needed to turn the basic samples into a sample from the distribution.

More concretely, with a default of `T = Float64` one would obtain
```julia
julia> typeof(rand(Gamma(0.5f0))) # = typeof(rand(Float64, Gamma(0.5f0)))
Float64

julia> typeof(rand(Gamma(0.5))) # = typeof(rand(Float64, Gamma(0.5)))
Float64

julia> typeof(rand(Gamma(big(0.5)))) # = typeof(rand(Float64, Gamma(big(0.5))))
BigFloat

julia> typeof(rand(Float32, Gamma(0.5f0)))
Float32

julia> typeof(rand(Float32, Gamma(0.5)))
Float64

julia> typeof(rand(Float32, Gamma(big(0.5))))
BigFloat

[...]

PR #101, opened 2021-06-01
LKJL measures

cscherrer:


src/parameterized/LKJL.jl
```diff
@@ -7,11 +7,11 @@ export LKJL
 The LKJ distribution (Lewandowski et al 2009) for the Cholesky factor L of correlation
 matrices.
 
-A correlation matrix ``Ω=LL'`` has the density ``|Ω|^{η-1}``. However, it is usually not
+A correlation matrix ``Ω=L*L'`` has the density ``|Ω|^{η-1}``. However, it is usually not
```
sethaxen, line 10:
```suggestion
A correlation matrix ``Ω=L*L'`` has a density proportional to ``|Ω|^{η-1}``. However, it is usually not
```

src/parameterized/LKJL.jl
```diff
@@ -7,11 +7,11 @@ export LKJL
...
 necessary to construct ``Ω``, so this distribution is formulated for the Cholesky
 decomposition `L*L'`, and takes `L` directly.
```
sethaxen, line 12:
```suggestion
decomposition `L*L'` and is supported on the lower triangular cholesky factor `L`.
```

src/parameterized/LKJL.jl
```diff
@@ -24,8 +24,20 @@ struct LKJL{k, N, T} <: ParameterizedMeasure{N}
...
+LKJL{k}(nt::NamedTuple{N,T}) where {k,N,T} = LKJL{k,N,T}(nt)
+
 LKJL(k,η) = LKJL{k, (:η,), Tuple{typeof(η)}}((η=η,))
 
+LKJL(k) = LKJL(k, 1.0)
+
+LKJL{k}(η) where {k} = LKJL{k, (:η,), Tuple{typeof(η)}}((η=η,))
+
+function Base.show(io::IO, d::LKJL{k}) where {k}
+    println(io, "LKJL{", k, "}(", getfield(d, :par), ")")
```
sethaxen, line 36:
```suggestion
    println(io, "LKJL{$k}(", getfield(d, :par), ")")
```

src/parameterized/LKJL.jl
```diff
@@ -24,8 +24,20 @@ struct LKJL{k, N, T} <: ParameterizedMeasure{N}
     par :: NamedTuple{N,T}
 end
 
+LKJL{k}(nt::NamedTuple{N,T}) where {k,N,T} = LKJL{k,N,T}(nt)
+
 LKJL(k,η) = LKJL{k, (:η,), Tuple{typeof(η)}}((η=η,))
```
sethaxen, line 29: With the below constructor, this can be simpler
```suggestion
LKJL(k,η) = LKJL{k}(η)
```

src/parameterized/LKJL.jl
```diff
@@ -24,8 +24,20 @@ struct LKJL{k, N, T} <: ParameterizedMeasure{N}
...
 
+LKJL(k) = LKJL(k, 1.0)
```
sethaxen, line 31: Perhaps it's better to restrict the type of `k` so a user is less like to put eta here, and use integer `1` for eta to avoid promoting to `Float64` in the log density if the user passes a matrix with an eltype like `Float32`:
```suggestion
LKJL(k::Integer) = LKJL(k, 1)
```

src/parameterized/LKJL.jl
```diff
@@ -38,17 +50,24 @@ function logdensity(d::LKJL{k}, L::Union{LinearAlgebra.AbstractTriangular, Diago
...
     @tullio s = (c - i) * log(L[i,i])
     return s
 end
 
 using TransformVariables
 
-TransformVariables.as(::LKJL{k}) where {k} = TransformVariables.CorrCholeskyFactor(k)
+TransformVariables.as(::LKJL{k}) where {k} = CorrCholeskyLower(k)
 
+# TODO: I think this is wrong
```
sethaxen, line 64: Not certain. Why do you think this is wrong?

src/parameterized/LKJL.jl
```diff
@@ -38,17 +50,24 @@ function logdensity(d::LKJL{k}, L::Union{LinearAlgebra.AbstractTriangular, Diago
...
 function basemeasure(μ::LKJL{k}) where {k}
     t = as(μ)
     d = dimension(t)
     return Pushforward(t, Lebesgue(ℝ)^d, false)
 end
+
+function Base.rand(rng::AbstractRNG, ::Type, d::LKJL{k}) where {k}
```
sethaxen, line 71: Note: this can be done without the `cholesky` decomposition, by randomly sampling the canonical partial correlations, as in the LKJ paper, and then mapping them to the cholesky factor instead of the correlation matrix. Stan [implements this](https://github.com/stan-dev/math/blob/develop/stan/math/prim/prob/lkj_corr_cholesky_rng.hpp). But that can be for a future PR.

The `::Type` parameter is not used. Should it be? Also, does it matter for MeasureTheory if the output type is `LowerTriangular` or a lower triangular `Matrix`?

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+#### correlation cholesky Lower
+####
+# Adapted from https://github.com/tpapp/TransformVariables.jl/blob/master/src/special_arrays.jl
+
+const TV = TransformVariables
+using LinearAlgebra
+
+"""
+    CorrCholeskyLower(n)
+Cholesky Lower of a correlation matrix of size `n`.
```
sethaxen, line 12: This phrasing is odd.
```suggestion

Cholesky Lower of a correlation matrix of size `n`.
```

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+Transforms ``n×(n-1)/2`` real numbers to an ``n×n`` lower-triangular matrix `L`, such that
```
sethaxen, line 13:
```suggestion
Transforms ``n(n-1)/2`` real numbers to an ``n×n`` lower-triangular matrix `L`, such that
```

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+`L*L'` is a correlation matrix (positive definite, with unit diagonal).
+# Notes
```
sethaxen, line 15:
```suggestion

# Notes
```

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+If
+- `z` is a vector of `n` IID standard normal variates,
+- `σ` is an `n`-element vector of standard deviations,
+- `L` is obtained from `CorrCholeskyLower(n)`,
+then `Diagonal(σ) * L * z` will be a multivariate normal with the given variances and
```
sethaxen, line 20:
```suggestion
then `Diagonal(σ) * L * z` is a zero-centered multivariate normal variate with the standard deviations `σ` and
```

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+correlation matrix `L * L'`.
+"""
+struct CorrCholeskyLower <: TV.VectorTransform
+    n::Int
+    function CorrCholeskyLower(n)
+        # @argcheck n ≥ 1 "Dimension should be positive."
+        new(n)
+    end
```
sethaxen, line 28: This is just the default constructor
```suggestion
```

src/transforms/corrcholeskylower.jl
```diff
@@ -0,0 +1,65 @@
...
+end
+
```
sethaxen, line 30: I think this can be greatly simplified (and avoid using non-API functions of TransformVariables) by just calling the corresponding functions for `CorrCholeskyUpper`:

```julia
TV.dimension(t::CorrCholeskyLower) = TV.dimension(TV.CorrCholeskyUpper(t.n))

function TV.transform_with(flag::TV.LogJacFlag, t::CorrCholeskyLower, x::AbstractVector, index)
    U, ℓ, index = TV.transform_with(flag, TV.CorrCholeskyUpper(t.n), x, index)
    return U', ℓ, index
end

TV.inverse_eltype(t::CorrCholeskyLower, L::LowerTriangular) = TV.extended_eltype(L)

function TV.inverse_at!(x::AbstractVector, index, t::CorrCholeskyLower, L::LowerTriangular)
    return TV.inverse_at!(x, index, TV.CorrCholeskyUpper(t.n), L')
end
```

sethaxen (changes requested):
Looks good! I had a number of suggestions. Also, it seems the normalization constant should appear somewhere, perhaps in `base_measure`? Distributions has an implementation of this that we could access.

Also this PR needs tests.

src/parameterized/LKJL.jl
cscherrer: This starts with Lebesgue measure in ℝⁿ, then stretches the space using the transform. I think what we end up with will be an artifact of our preference to use ℝⁿ, instead of whatever is natural for the space. There must be an easy way to check this sort of thing, but I haven't thought of it yet.

src/parameterized/LKJL.jl
cscherrer: I'm not entirely happy with our random generation, so I'm open to ideas in general. But this `::Type` setup is one aspect I think is very nice. 

This comes from @devmotion in https://github.com/cscherrer/MeasureTheory.jl/issues/51; the idea is that every `rand` call will ententually (after a series of dispatches) call at least one `rand(::AbstractRNG, T::Type)`. The resulting `::T` will the propagate back up to the final result.

cscherrer: > Also, does it matter for MeasureTheory if the output type is LowerTriangular or a lower triangular Matrix?

I'm not sure. I haven't dug into the internals of `LowerTriangular`, but here's one thing that's pretty interesting:
```julia
julia> X = randn(4,4)
4×4 Matrix{Float64}:
  0.402848  0.787592  -0.110378   0.669417
  0.183548  0.48167    0.8996    -0.616217
 -0.108725  1.62248   -1.0993    -0.143815
 -0.305961  0.341102   1.95539    0.849061

julia> sizeof(X)
128

julia> using LinearAlgebra

julia> sizeof(LowerTriangular(X))
8
```

This seems like a big advantage, but strangely I don't see a way to construct one directly (only to build a Matrix and then convert it). What would you suggest?

src/transforms/corrcholeskylower.jl
cscherrer: Haha, that's my search and replace showing. It had been "Cholesky Factor" :)

src/transforms/corrcholeskylower.jl
cscherrer: Ah, right. I guess Tamas had this to have the `@argcheck` available. I'd love a nice way to have optional arg checks. There's
https://github.com/simeonschaub/OptionalArgChecks.jl
but it says it broken. Let's drop this for now and come back to the question of argchecks once we have more of the functionality in place

src/transforms/corrcholeskylower.jl
cscherrer: Great idea!

src/parameterized/LKJL.jl
sethaxen: > The resulting ::T will the propagate back up to the final result.

I see, so the reason `::Type` is unused here is because Distributions's `rand` function doesn't accept this type parameter, correct? Okay, what was confusing me was that it was there but unused.

> I'm not sure. I haven't dug into the internals of `LowerTriangular`, but here's one thing that's pretty interesting:

Weird. I have no idea what `sizeof` does, and it just redirects to a `Core.sizeof` function, which `@which` does not work for, saying its not a generic function, so I don't know what this even means.

`LowerTriangular` has a simple implementation though. It's just a wrapper around another matrix for storage, where it only considers the lower triangle of that matrix. So in your example, `LowerTriangular` just wraps your original `X`; it is no more or less space-efficient than `X`. Potentially by wrapping you could leave the strict upper triangle with undefined entries, but not even functions in LinearAlgebra seem to do that.

src/parameterized/LKJL.jl
sethaxen: Oh hm, then yes, I think this is not quite correct. Because would this not also imply that one could evaluate `logdensity(Lebesgue(ℝ)^d, L)` for lower triangular `L`, i.e. that our samples are elements of `ℝ^d`?

Otherwise, wouldn't the base measure be `Lebesgue(ℝ)^(k,k)`, or however that is written, with the constant normalization factor?

src/parameterized/LKJL.jl
cscherrer: > I see, so the reason `::Type` is unused here is because Distributions's `rand` function doesn't accept this type parameter, correct?

Yep, that's it. David had a PR to add this to Distributions, but I'm not sure of the status.

> Weird. I have no idea what sizeof does, and it just redirects to a Core.sizeof function, which @which does not work for, saying its not a generic function, so I don't know what this even means.

It's suposed to be the size in bytes. But looking again I think there must be a bug for this case.

> LowerTriangular has a simple implementation though. It's just a wrapper around another matrix for storage, where it only considers the lower triangle of that matrix. So in your example, LowerTriangular just wraps your original X; it is no more or less space-efficient than X. Potentially by wrapping you could leave the strict upper triangle with undefined entries, but not even functions in LinearAlgebra seem to do that.

Ah, ok. In that case I think we do want `LowerTriangular`. We can allocate the matrix as `undef`, write only the values we need, and the constructor will "mask" the remaining `undef`s

src/parameterized/LKJL.jl
cscherrer: I think it's more complicated than that, since we need to match the dimension. For example,
```julia
julia> d = LKJL(3, 2.0)
LKJL{3}((η = 2.0,))


julia> rand(d)
3×3 LowerTriangular{Float64, Matrix{Float64}}:
  1.0         ⋅         ⋅ 
  0.232191   0.97267    ⋅ 
 -0.489373  -0.595999  0.636631
```
This looks like it has 5 degrees of freedom, but in reality,

```julia
julia> dimension(as(d))
3
```

cscherrer: `Pushforward` looks like this:
```julia
struct Pushforward{F,M} <: AbstractMeasure
    f::F
    μ::M
    logjac::Bool
end
```

It might work to use this transform with `logjac=false`. I haven't tried that before, but... maybe?

cscherrer: Oh, that's what I have :laughing: 

Ok maybe this is right. I guess the check is that the logdensity of the base measure should be constant on tranformed values.

sethaxen:
I haven't had a chance to look at the recent comments/changes, but I also wanted to point out that in https://github.com/JuliaStats/Distributions.jl/issues/1336 we're discussing the possibility of implementing something like an `LKJCholesky` distribution whose support are objects of type `Cholesky` from LinearAlgebra. This has the benefit of not committing to either upper or lower triangular parameterizations.

That might be something to consider here or in a future PR. So e.g. in Soss, a user might have
```julia
Ω ~ LKJCholesky(η=2)
y ~ MvNormal(Ω=Ω, σ=σ)
```
where the `logdensity` function for `MvNormal` would have a term like `z = (x .- μ) ./ σ; return -(dot(z, Ω \ z) + logdet(Ω)) / 2`. This would work for matrix- or `Cholesky`-valued `Ω`, so no special overloads are needed for `Cholesky`. As a bonus, you can decide whether you want `Cholesky` to store the upper or lower triangular factor (it has an `uplo` field for this), so you could even choose upper and use `CorrCholeskyUpper` directly instead of wrapping it. What do you think?

cscherrer:
That's really interesting. It sounds like a good idea. I'll think about it some more and also check out the discussion. Thanks!

cscherrer:
Thinking some more about what this would mean for PPL. Just thinking out loud here...

**`logdensity`**

It seems like using this form with a `logdensity` would require first constructing the `Cholesky`, which takes some extra work. But that may not really be the case; we could add methods for `LowerTriangular` and `UpperTriangular` that just transpose to get `U` or `L`, respectively.

**`rand`**

Currently we only write into the lower triangular. The upper portion is `undef`, and `LowerTriangular` hides it.

I had assumed `cholesky` works by putting everything into a single matrix, then getting `L` and `U` by hiding the upper or lower portion. But that's not right! Check it out:

```julia
julia> A = [4. 12. -16.; 12. 37. -43.; -16. -43. 98.]
3×3 Matrix{Float64}:
   4.0   12.0  -16.0
  12.0   37.0  -43.0
 -16.0  -43.0   98.0

julia> C = cholesky(A)
Cholesky{Float64, Matrix{Float64}}
U factor:
3×3 UpperTriangular{Float64, Matrix{Float64}}:
 2.0  6.0  -8.0
  ⋅   1.0   5.0
  ⋅    ⋅    3.0

julia> C.L.data
3×3 Matrix{Float64}:
  2.0  12.0  -16.0
  6.0   1.0  -43.0
 -8.0   5.0    3.0

julia> C.U.data
3×3 Matrix{Float64}:
   2.0    6.0  -8.0
  12.0    1.0   5.0
 -16.0  -43.0   3.0
```

Anyway, whether we took the `cholesky` approach or just used a single "parent" matrix, we'd end up with roughly twice the memory writes, which is not great. So I think it's good to have upper and lower available separately, even if we also add a version that contains both.

sethaxen:
> I had assumed `cholesky` works by putting everything into a single matrix, then getting `L` and `U` by hiding the upper or lower portion. But that's not right!

For this, it's constructive to check the [implementation of `getproperty` for `Cholesky`](https://github.com/JuliaLang/julia/blob/v1.6.1/stdlib/LinearAlgebra/src/cholesky.jl#L445-L457). `Cholesky` has an `uplo` field that specifies whether the upper triangle or the lower triangle is used to store the upper (or lower) Cholesky factor. `getproperty` checks this field. if `uplo` is `L`, then it just wraps the stored `factors` with `LowerTriangular` and returns it with no copy (if the upper factor is requested, then it copies to keep type-stability). If `uplo` is `U`, then the reverse is true. All `Cholesky`-involved code checks the `uplo` field. Here's the odd thing: `cholesky` does not offer an option to choose whether `uplo` is `U` or `L`. It's always `U`. We can test this:

```julia
julia> F = cholesky(A)

julia> FU = Cholesky(F.U.data, 'U', 0);

julia> @allocated FU.factors
32

julia> @allocated FU.U.data
32

julia> @allocated FU.L.data
192

julia> FL = Cholesky(F.L.data, 'L', 0);

julia> @allocated FL.factors
32

julia> @allocated FL.U.data
192

julia> @allocated FL.L.data
32
```

Since the `MvNormal` example I gave dispatches to overloads that do the right thing for both `uplo` values, the case to consider is when the user intends to access the cholesky factor directly. The most important case here is when the user wants a non-centered parameterization. In this case, I think they are more likely to want the `L` factor than the `U`, so choosing `uplo='L'` would avoid the copy. The only downside is that, if the user knew they were being given the `L` factor but wanted `U`, they could adjoint with no copy. One solution is to add an optional `uplo` entry to the distribution, which determines whether the upper or lower factor is used.

> Anyway, whether we took the `cholesky` approach or just used a single "parent" matrix, we'd end up with roughly twice the memory writes, which is not great. So I think it's good to have upper and lower available separately, even if we also add a version that contains both.

I think the `LKJCholesky` approach, where the measure has an optional `uplo` field or type parameter, would handle these cases. It has the benefit of returning a native Julia type and enabling the user to use the `Cholesky` decomposition directly when that's what they want. It also makes it simple to use in the `MvNormal` distribution. Its main downside is that the user might need to think about the storage of `Cholesky` and set a type parameter or field to get the storage they want with no copy. But I do think most users would not need to think about this.

A final point against a `Cholesky` return value (or really, any return value that is not an `AbstractArray` or `Number`), is that it makes it harder to compute diagnostics. i.e. to compute Rhat, you'd need to extract the desired cholesky factor. But I'm guessing MeasureTheory and Soss both want the flexibility of having arbitrary types be parameters, right?

cscherrer:
Oh, I see. So a `::Cholesky` has three fields: `L`, `U`, and `UL`. `UL` seems to be the "native" representation (either `UpperTriangular` or `LowerTriangular`), and `L` and `U` give you flexibility, though one of those will require extra computation.

> A final point against a Cholesky return value (or really, any return value that is not an AbstractArray or Number), is that it makes it harder to compute diagnostics. i.e. to compute Rhat, you'd need to extract the desired cholesky factor. But I'm guessing MeasureTheory and Soss both want the flexibility of having arbitrary types be parameters, right?

I don't see how this complicates Rhat, but yes we definitely want flexibility.

sethaxen:
> Oh, I see. So a `::Cholesky` has three fields: `L`, `U`, and `UL`. `UL` seems to be the "native" representation (either `UpperTriangular` or `LowerTriangular`), and `L` and `U` give you flexibility, though one of those will require extra computation.

Almost, it does have three fields, but they're `factors`, `uplo`, and `info`. `info` is always 0 or 1 (1 iff the cholesky decomposition failed). `uplo` we've discussed. `factors` is the storage. In addition to the fields, it has 3 convenient properties: `L`, `U`, and `UL`. `L` and `U` are type-stable copies or views of `factors` wrapped in the corresponding triangular type, while `UL` is the type-unstable "give me whichever does not copy" property (this is only used for `show`, I think).

> I don't see how this complicates Rhat, but yes we definitely want flexibility.

Diagnostics are usually computed/reported on the constrained variables. Which means we need to be able to compute variances. For matrix distributions, each element is treated separately. But a `Cholesky` object is only matrix-like; it is not an `AbstractMatrix`. So it does not support the necessary operations to compute elementwise variance, and one would probably need to convert it to some `Matrix` for diagnostics.

cscherrer:
> > Oh, I see. So a `::Cholesky` has three fields: `L`, `U`, and `UL`. `UL` seems to be the "native" representation (either `UpperTriangular` or `LowerTriangular`), and `L` and `U` give you flexibility, though one of those will require extra computation.
> 
> Almost, it does have three fields, but they're `factors`, `uplo`, and `info`. `info` is always 0 or 1 (1 iff the cholesky decomposition failed). `uplo` we've discussed. `factors` is the storage. In addition to the fields, it has 3 convenient properties: `L`, `U`, and `UL`. `L` and `U` are type-stable copies or views of `factors` wrapped in the corresponding triangular type, while `UL` is the type-unstable "give me whichever does not copy" property (this is only used for `show`, I think).

Ah, right. I was thinking properties but saying fields :)
 
> Diagnostics are usually computed/reported on the constrained variables. Which means we need to be able to compute variances. For matrix distributions, each element is treated separately. But a `Cholesky` object is only matrix-like; it is not an `AbstractMatrix`. So it does not support the necessary operations to compute elementwise variance, and one would probably need to convert it to some `Matrix` for diagnostics.

Could this computed in the "raw" form, before the transformation?

cscherrer:
I've added [lkj-cholesky.jl](https://github.com/cscherrer/MeasureTheory.jl/pull/101/files#diff-5482d13f7ce964d626b2cb17f821bd279075bee0de21c955268c212da8dbbd20), and I'm thinking of dropping `LKJL` and using this in its place. I've also switched to `cholesky(Positive, ...)` from [PositiveFactorizations.jl](https://github.com/timholy/PositiveFactorizations.jl). This claims to have some stability advantages, and it's also nice that it works in terms of the `L` component. This is really only for `rand`, and I think we won't need it at all once we get to sampling the lower-diagonal directly.

I'll get some tests set up for this, then we can look at whether to transition entirely to this approach.

sethaxen:
> I've added [lkj-cholesky.jl](https://github.com/cscherrer/MeasureTheory.jl/pull/101/files#diff-5482d13f7ce964d626b2cb17f821bd279075bee0de21c955268c212da8dbbd20), and I'm thinking of dropping `LKJL` and using this in its place. I've also switched to `cholesky(Positive, ...)` from [PositiveFactorizations.jl](https://github.com/timholy/PositiveFactorizations.jl). This claims to have some stability advantages, and it's also nice that it works in terms of the `L` component. This is really only for `rand`, and I think we won't need it at all once we get to sampling the lower-diagonal directly.

Interesting, I don't really know anything about PositiveFactorizations.jl. If nothing else, it's nice that this doesn't just raise an error of the matrix is slightly not positive-definite. I plan to implement random sampling as part of https://github.com/JuliaStats/Distributions.jl/issues/1336, and then I'll add a PR here to use that sampling.

> I'll get some tests set up for this, then we can look at whether to transition entirely to this approach.

Sounds good!

sethaxen:
Looks like `LKJCholesky` is just a placeholder for now. Let me know when it's ready for review.

cscherrer:
Great! Ok, two questions:
1. Do you think we should drop LKJL entirely in favor of LKJCholesky?
2. It's really common to use this as a prior with a constant `eta`, in which case the normalizing constant (even the part that depends on `eta`) just gets in the way. Currently, base measures depend only on the type, and not on the values of parameters. But I'm thinking of changing that for cases like this. Hmm, maybe this one isn't so much a question, just a thing I've been thinking about.

sethaxen:
> 1. Do you think we should drop LKJL entirely in favor of LKJCholesky?

I think so. Its functionality is contained within LKJCholesky, which is more general.

> 2. It's really common to use this as a prior with a constant `eta`, in which case the normalizing constant (even the part that depends on `eta`) just gets in the way. Currently, base measures depend only on the type, and not on the values of parameters. But I'm thinking of changing that for cases like this.

How would you detect if eta is constant or not?

cscherrer:
> I think so. Its functionality is contained within LKJCholesky, which is more general.

:+1: 

> How would you detect if eta is constant or not?

You nailed it, this is exactly the question :) 

Let's say `k` is fixed, so we just have a measure over Cholesky decompositions that's parameterized by `eta`. For a given `y`, we get the density at that `y` of the measure relative to some base measure.

For the base measure, I think we can just do the same thing, it's parameterized by `eta`, and it just happens to not depend on `y` wrt Lebesgue measure. Then, if you want to have a prior on `eta`, you need to drill down far enough that the base measure doesn't have `eta` as a parameter.

cscherrer:
Ok @sethaxen, I think it's ready for review. Let's merge this once everything but the base measure is correct (but not tag a release yet), then I'll have another PR to try the parameterized base measure business.

PR #105, opened 2021-06-09
LogLikelihood update

cscherrer:


src/combinators/likelihood.jl
```diff
@@ -1,15 +1,142 @@
...
+
+Likelihoods are most commonly used in conjunction with an existing _prior_
+measure to yield a new measure, the _posterior_. In Bayes's Law, we have
+
+``P(θ|x) ∝ P(θ) P(x|θ)``
+
+Here ``P(θ)`` is the prior. If we consider ``P(x|θ)`` as a function on ``θ``,
+then it is called a likelihood.
+
+Since measures are mostly commonly manipulated using `density` and `logdensity`,
```
nignatiadis, line 16:
```suggestion
Since measures are most commonly manipulated using `density` and `logdensity`,
```

src/combinators/likelihood.jl
```diff
@@ -1,15 +1,142 @@
...
+    -1.0986122886681098
+
+-----------------------
+
+Finally, let's return to the expression for Bayes's Law, 
+
+``P(θ|x) ∝ P(θ) P(x|θ)``
+
+The product on the right side is computed pointwise. To work with this in
+MeaureTheory, we have a "pointwise product" `⊙`, which takes a measure and a
```
nignatiadis, line 87:
```suggestion
MeasureTheory, we have a "pointwise product" `⊙`, which takes a measure and a
```

src/combinators/likelihood.jl
```diff
@@ -1,15 +1,142 @@
...
+likelihood, and returns a new measure.
```
nignatiadis, line 88:
```suggestion
likelihood, and returns a new measure, that is, the unnormalized posterior that has density ``P(θ) P(x|θ)`` with respect to the base measure of the prior.
```

src/combinators/likelihood.jl
```diff
@@ -1,15 +1,142 @@
...
+
+function Likelihood(μ::M, constraint::NamedTuple, x) where {M<:AbstractMeasure}
+    Likelihood((M, constraint), x)
+end
+
+function (ℓ::Likelihood{Tuple{M,NT}})(p::NamedTuple) where {M<:Type, NT <: NamedTuple}
+    (D, constraint) = ℓ.f
+    return logdensity(D(merge(p, constraint)), ℓ.x)
+end
+
```
nignatiadis, line 132: I think it's not documented right now that `Likelihood` objects can act as callables (and specifically that they evaluated the loglikelihood).

nignatiadis (commented):
Looks great to me! Only left minor remarks about the docs.

src/combinators/likelihood.jl
cscherrer: Ah, right. I was thinking maybe we should remove this, since it's kind of confusing whether the result would give the density or log-density. The ambiguity is not great. What do you think?

cscherrer: I updated it so we define `logdensity` directly, and drop callable likelihoods. We can consider adding them back in later if it's awkward without this.

PR #110, opened 2021-06-25
Dev

cscherrer:
- Added dependencies for ConstructionBase, FillArrays, and Tricks
- Moved to using `Realized` for instantiations of many measures. This reduces allocation, and can be `collect`ed if we like
- Lots of code reduction for `For`
- Changed `ProductMeasure` to have `f` and `par`. `par` is a collection of parameters, and `f` is a function to apply to each entry
- Changed `Likelihood` so the first argument is a `Kernel`
- Added a `Const` constructor for representing constant functions (funny this isn't in Base that I know of)
- `𝒹` and `∫` in log space are now written `log𝒹` and `∫exp`
- Update `kernel` so the constructor is explicit, for cases where it's not type-level
-


src/const.jl
```diff
@@ -0,0 +1,11 @@
+struct Const{T} <: Function
```
phipsgabler, line 1: As I said on Zulip, this exists as `Returns` in [1.7](https://github.com/JuliaLang/julia/pull/39794).  Maybe use some compat tricks to use that instead.

Also, the [constructor](https://github.com/JuliaLang/julia/pull/39794/files#diff-12e7a6522633012a408b1bdee7639e8cb722617fe1a8ed6a3881bf4ad1ebdbbdR941) over there is different, maybe that's relevant.

cscherrer: Hey @phipsgabler, sorry I had missed this. I agree, let's move toward `Returns` and `Compat`

PR #112, opened 2021-07-05
remove test failure if warnings are printed when package is imported

KristofferC:
This is a very brittle test, if any of the dependencies shows something to stderr this will fail. On 1.7 this happens (https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_hash/9bb1633_vs_dd12291/MeasureTheory.1.7.0-beta2-9a791aa09fb.log). For example:

```
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead
```

I would say this test needs to be more precise if it is to be useful.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -5,14 +5,7 @@ using Random
 using LinearAlgebra
 using DynamicIterators: trace, TimeLift
 using TransformVariables: transform, as𝕀, inverse
-
-if Base.VERSION ≥ v"1.6"
-    @testset "No warnings on import" begin
-        @test_nowarn @eval using MeasureTheory
-    end
-else
-    @eval using MeasureTheory
-end
+using MeasureTheory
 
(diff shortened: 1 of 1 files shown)


cscherrer:
Thanks @KristofferC . We were having trouble with this for a while because of some subtleties in KeywordCalls.jl. I think we're past this now, so I'll merge this to be sure we're ok for 1.7. If we start seeing issues like this again we can revisit and try to be more precise as you suggest.

DilumAluthge:
@cscherrer We should probably register a new release, since PkgEval only uses registered versions of packages

cscherrer:
Sounds good. I've got some changes coming in `dev`, but we can push this to the front

cscherrer:
I updated the version to 0.8.2. Let's make sure tests pass, then we can register

PR #116, opened 2021-07-13
Kusti's NegativeBinomial update

cscherrer:
Updated from https://github.com/cscherrer/MeasureTheory.jl/pull/115
cc @kskyten


cscherrer:
Thanks @kskyten ! I made some edits, with the most significant being that Distributions.jl parameterizes `NegativeBinomial` in terms of `r` and `p`, not `r` and `λ`. I think we can merge this once tests pass

PR #122, opened 2021-07-28
draft MvNormal

cscherrer:


mschauer:
What do you think of going the way of GaussianDistributions.jl and have one object for 1d and nd Gaussians?

cscherrer:
> What do you think of going the way of GaussianDistributions.jl and have one object for 1d and nd Gaussians?

It's an interesting idea, my only concerns would be 
1. The types are different - scalar vs (possibly 1d) vector
2. It could be a little confusing coming from Distributions.

Then again, I've strayed a little from Distributions's naming already; `TDist` is an awful name IMO, so in MeasureTheory it's `StudentT`

PR #126, opened 2021-08-03
TransformVariables => Bijectors

cscherrer:
Let's see if we can transition to Bijectors.jl. This PR will depend on
TuringLang/Bijectors.jl#183

cc @torfjelde


cscherrer:
Closing this, since https://github.com/TuringLang/Bijectors.jl/pull/183 is outdated now anyway. If we want to come back to this, it should be a package extension, either for MeasureTheory or MeasureBase.

PR #130, opened 2021-08-06
Adds mass and normalize for a measure

### LINKED ISSUE #128: Normalizing a measure
In https://github.com/JuliaManifolds/ManifoldMeasures.jl/issues/7 we were discussing that we need a way to map an (unnormalized) Hausdorff measure to the corresponding probability measure by normalizing. This seems general enough that it could be defined here.

What about something like this?
```julia
LinearAlgebra.normalize(μ::AbstractMeasure) = inv(total_mass(μ)) * μ
unit() = true  # just a function to use for dispatch.
total_mass(μ::AbstractMeasure) = ∫(unit, μ)
total_mass(μ::AbstractWeightedMeasure) = exp(μ.logweight) * total_mass(μ.base)
is_probability_measure(μ::AbstractMeasure) = isone(total_mass(μ))
```

Perhaps there is a more measure-theoretic term than "normalization", and we should not then overload `normalize`. In the case of a Hausdorff measure, `total_mass` would compute the volume/area of the manifold in some embedded metric space.

It might be preferable to define `log_total_mass` instead.

sethaxen:
Fixes #128

### DIFF
--- a/src/MeasureTheory.jl
+++ b/src/MeasureTheory.jl
@@ -30,6 +30,8 @@ export basemeasure
 
 export as
 
+export mass, logmass, is_probability_measure, normalize
+
 abstract type AbstractMeasure end
 
 sampletype(μ::AbstractMeasure) = typeof(testvalue(μ))
@@ -54,6 +56,7 @@ include("absolutecontinuity.jl")
 include("parameterized.jl")
 include("macros.jl")
 include("resettablerng.jl")
+include("mass.jl")
 
 include("primitive.jl")
 include("primitives/counting.jl")
--- a/src/combinators/weighted.jl
+++ b/src/combinators/weighted.jl
@@ -20,6 +20,8 @@ function logdensity(sm::AbstractWeightedMeasure, x)
     logdensity(sm.base, x) + sm.logweight
 end
 
+logmass(d::AbstractWeightedMeasure) = logmass(d.base) + d.logweight
+
 ###############################################################################
 
 struct WeightedMeasure{R,M} <: AbstractWeightedMeasure
--- a/src/distributions.jl
+++ b/src/distributions.jl
@@ -35,6 +35,10 @@ end
 
 ∫(::typeof(identity), ::Dists.Distribution) = 1.0
 
+# all Distributions distributions are probability measures
+logmass(μ::DistributionMeasure) = true
+is_probability_measure(μ::DistributionMeasure; kwargs...) = true
+
 logdensity(μ::Dists.Distribution, x) = Dists.logpdf(μ,x)
 Dists.logpdf(d::AbstractMeasure, x) = logdensity(d,x)
 
--- a/src/mass.jl
+++ b/src/mass.jl
@@ -1,0 +1,40 @@
+using LinearAlgebra: LinearAlgebra
+
+"""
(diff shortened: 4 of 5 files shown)


gdalle:
Hey there @sethaxen! I wanted to do that too in order to simulate Poisson processes on a given set, but I was going to make the relevant measures callable: something like
```julia
(::Lebesgue{ℝ})(interval::IntervalSets.AbstractInterval) = duration(interval)
```
Is that compatible with your approach? If not, which do you think is best?

gdalle:
After some more thinking, I feel like it the notion of total mass is intimately linked with the subset we consider. If we want to integrate (numerically) a measure on a set, do we define the measure on that set (as in `Lebesgue(S)`) or do we apply the measure to that set (as in `(::Lebesgue)(S)`) ? I think the second one makes more sense, but I would like @cscherrer's input

sethaxen:
The purpose of this PR was to provide an easy way to normalize a measure, but I don't think this is the right approach. For example, suppose we have a measure `μ` that we want to normalize, and we know how to sample from the normalized measure, but we don't know the total mass. Then this approach will not work, since it relies on computing the total mass.

I've been experimenting with a `Normalized` measure that simply represents the normalized version of whatever measure it wraps: https://github.com/JuliaManifolds/ManifoldMeasures.jl/blob/main/src/normalized.jl You can see, for example, how it is used here: https://github.com/JuliaManifolds/ManifoldMeasures.jl/blob/55f1713df5257c0723c6f3ff09d51e985777bb34/src/hausdorff.jl#L140-L150. This allows us to avoid computing the `logmass` unless absolutely necessary.

> After some more thinking, I feel like it the notion of total mass is intimately linked with the subset we consider. If we want to integrate (numerically) a measure on a set, do we define the measure on that set (as in `Lebesgue(S)`) or do we apply the measure to that set (as in `(::Lebesgue)(S)`) ? I think the second one makes more sense, but I would like @cscherrer's input

I completely agree that for computing mass in general, we need a generic way to define sets, and it would be nice to be able to flexible hook into numerical integration packages for that. But because something like `logmass` could be called deep in some user code, we need to find a way for that user to specify parameters of the numerical integration algorithm.

cscherrer:
Yes, I think these are two different things. For a measure `μ`, I think it makes sense to ask for `μ(S)` for some set `S`. For a point process, `S` might be a `Set` (if we have exchangability), or maybe an iterable. 

Otherwise, we need to be a little restrictive, since things can get out of hand. I do think there are some things we can do for manifolds or measures dominated by Lebesgue measure, moslty in terms of intervals. 

[IntervalSets](https://github.com/JuliaMath/IntervalSets.jl) is very lightweight, so I think we could depend on that. Beyond that, there's lots of great stuff in [JuliaIntervals](https://github.com/JuliaIntervals) that we could get to through an extension package. Those (at least mostly) don't depend on IntervalSets, but a conversion should be very easy.

gdalle:
So where does that leave us? I'm a bit lost as to what to implement

cscherrer:
> So where does that leave us? I'm a bit lost as to what to implement

Sounds like we collectively need to get in sync. I'll write a summary on Zulip, and we can talk some more about a broad view of where we're heading.

PR #133, opened 2021-08-11
Multi-dimensional Lebesgue measure

gdalle:
Hey there!
As discussed, here's a proposed implementation of the domain representing $R^d$ as well as the Lebesgue measure that goes with it. For clarity's sake I would love to turn `EuclideanSpace{D}` into `\bbR{D}` but I don't know if it would still be possible to define a default `\bbR = \bbR{1}`.
What do you think?


cscherrer (commented):
Thanks for the PR! I like the idea of `EuclideanSpace`. Currently we'd instead use `Lebesgue(\bbR) ^ n`, which can be useful for some constructions, but for others it could be more natural to define this directly as you do here.

For low-ish dimensions an `SVector` can work very well, but unfortunately this doesn't scale well. Also, it adds a `StaticArrays` dependency. I've heard from a few people a strong preference for keeping dependencies minimal. So I think I'd suggest...

1. We can add [Static.jl](https://github.com/SciML/Static.jl) as a dependency, and possibly [ArrayInterface](https://github.com/JuliaArrays/ArrayInterface.jl) if we need it.
2. `EuclideanSpace` can be defined as 
```julia
struct EuclideanSpace{T} <: AbstractDomain
    dimension :: T
end
```
3. For a `EuclideanSpace{Int}`, you'll need to check the `dimension` field, but we'll also have `EuclideanSpace{StaticInt{k}} where {k}` available for the static case.
4. If you have ideas for specialized methods for StaticArrays, we can add those in MultivariateMeasures.

Make sense?

mschauer:
I like
```
struct EuclideanSpace{T} <: AbstractDomain
    dimension :: T
end
```
because so straight forward and account for the static case with `StaticInt` if that works nicely.

gdalle:
Alright, I'm not sure I understand all the subtleties of the static integer but I'll implement this!

gdalle:
Also, why does CI keep failing me?

cscherrer:
I tightened the CI down a bit. The problem is here: https://github.com/cscherrer/MeasureTheory.jl/pull/133/checks?check_run_id=3310734415#step:6:148

So the issue is that `Static` has dependencies that don't allow for old versions of Julia. I set up [Aqua.jl](https://github.com/JuliaTesting/Aqua.jl) (great package) to test for this sort of thing. 

I'm not sure yet, but I think this may be an issue with Static.jl. We probably need to ask in an issue there if they have a workaround.

cscherrer:
@gdalle as for the static integer thing....
- [Here](https://juliaarrays.github.io/ArrayInterface.jl/dev/) are the docs for `ArrayInterface`, an effort to standardize this sort of thing
- [Here](https://github.com/cscherrer/MultivariateMeasures.jl/blob/main/src/parameterized/mvnormal.jl) is an example of using this in MultivariateMeasures. The `KnownSize` thing will likely change at some point.
- [Here](https://github.com/JuliaArrays/ArrayInterface.jl/issues/182) is some discussion that helped me better understand the approach

gdalle:
Replicated in #147

PR #137, opened 2021-08-13
Testvalues

cscherrer:


test/runtests.jl
```diff
@@ -21,6 +21,53 @@ function draw2(μ)
...
+    Uniform()
+    # CountingMeasure(Float64)
+    Dirac(π)
+    Lebesgue(ℝ)
+    # TrivialMeasure()
+]
+
+@testset "testvalue" begin
+    for μ in measures
+        @test logdensity(μ, testvalue(μ)) isa Float64
```
sethaxen, line 61: If this is a universal requirement, can you also document somewhere, perhaps on the definition of `testvalue` that it should produce a point from the sample space of the measure such that evaluating `logdensity` on it produces a `Float64`?

Also, if this is the beginning of a test suite that all user-implemented measures should follow, then perhaps you should create a `test_measure` function in the source code of this package so that dependents of this package can use that function, and for now, the tests will just include this line.

test/runtests.jl
```diff
@@ -21,6 +21,53 @@ function draw2(μ)
...
+    end
+    
+    @testset "testvalue(::Chain)" begin
+        mc =  Chain(x -> Normal(μ=x), Normal(μ=0.0))
+        r = testvalue(mc)
+        @test logdensity(mc, Iterators.take(r, 10)) isa Float64
```
sethaxen, line 67: This is a little unintuitive for me. So `testvalue`, which is I think a point on the sample space of `Chain`, cannot be passed to the `logdensity` function without some processing. Would it make more sense for `Chain` to be a family of measures that must be conditioned on a fixed number of steps (here, 10), in order to uniquely define a measure?

test/runtests.jl
```diff
@@ -21,6 +21,53 @@ function draw2(μ)
...
+    # transforms
+    3 * Normal()
+    Bernoulli(0.2)
+    Beta(2,3)
+    Binomial(10,0.3)
+    Cauchy()
+    Dirichlet(ones(3))
+    Exponential()
+    Gumbel()
+    # InverseGamma(2)
```
sethaxen, line 43: Are these commented out because they do not currently work? If so, perhaps better to put them in a different list and then apply `@test_broken`, so that the test suite tells us when they work and we need to move them from one list to the other?

test/runtests.jl
```diff
@@ -21,6 +21,53 @@ function draw2(μ)
...
+    Multinomial(n=10,p=[0.2,0.3,0.5])
+    # MvNormal
+    NegativeBinomial(5,0.2)
+    Normal(2,3)
+    Poisson(3.1)
+    StudentT(ν=2.1)    
+    Uniform()
+    # CountingMeasure(Float64)
+    Dirac(π)
+    Lebesgue(ℝ)
```
sethaxen, line 55: What about `Lebesgue` with some other supports?

test/runtests.jl
sethaxen: It would be useful to test somehow that `logdensity` does not needlessly promote. e.g. if the measure's parameters have a real eltype of `Float32`, and so does the test value, then `logdensity` should not promote to `Float64`.

cscherrer: > If this is a universal requirement, can you also document somewhere, perhaps on the definition of `testvalue` that it should produce a point from the sample space of the measure such that evaluating `logdensity` on it produces a `Float64`?

Interesting, I hadn't thought of it as a universal requirement. The point is that `testvalue` returns a value (no error) and the result can be evaluated with `logdensity`. But maybe it should be? Seems reasonable to ask that everything is `Float64` unless otherwise specified.

> Also, if this is the beginning of a test suite that all user-implemented measures should follow, then perhaps you should create a `test_measure` function in the source code of this package so that dependents of this package can use that function, and for now, the tests will just include this line.

Good idea, I'll add this.

> It would be useful to test somehow that `logdensity` does not needlessly promote. e.g. if the measure's parameters have a real eltype of `Float32`, and so does the test value, then `logdensity` should not promote to `Float64`.

Makes sense to me. Maybe `test_measure` chould have a type parameter, or loop through some floating-point types

test/runtests.jl
cscherrer: I agree it's unintuitive, but I think it's more a difficulty of the math than a problem with the implementation. It's a little like working with irrational numbers. Everything is well-defined, but you can never look at or do computations with the full set of decimal digits. Or even better a Gaussian process - the GP itself is a continuous function, but we can only ever evaluate it on a finite set of values.

test/runtests.jl
cscherrer: Yes, I agree these should be `@test_broken` unless fixed.

test/runtests.jl
cscherrer: There's not yet anything else that works. For higher dimensions, we currently do `Lebesgue(ℝ) ^ n`, though I think I like @gdalle 's idea of `EuclideanSpace` better.

Also, the whole `AbstractDomain` thing was just to get things going, and to avoid committing to a particular representation like `DomainsSets` that would bring additional dependencies. So maybe `AbstractDomain` is fine and just needs some more spaces added, or we could rethink how this is set up. In any case I'm open to feedback on the design.

gdalle: Btw, I had doubts about the `testvalue` of a multidimensional `Lebesgue` measure. The idea behind using `StaticVector`s was that their size was encoded in the type: otherwise, all multi-dimensional `Lebesgue` measures will have the same `testvalue` type. Is that a problem?

PR #142, opened 2021-08-16
migrate to MeasureBase

cscherrer:


cscherrer:
Currently seeing this issue:
```julia
Normal: Error During Test at /home/chad/tmp/MeasureTheory.jl/test/runtests.jl:128
  Got exception outside of a @test
  MethodError: no method matching asparams(::Type{Normal{(:μ, :σ)}}, ::Val{:μ})
```

This is because moving `@μσ_methods` to MeasureBase requires dropping the `asparams` methods that were previously part of the macro. But I'm wondering if we should have something more structural for this anyway, e.g. an `Affine` combinator that we could also use for nonstandard uniform, etc.

PR #146, opened 2021-08-20
Define Gamma parametrized measure

gdalle:
I tried to define a Gamma measure by copying and adapting the code from the Beta measure. Does that look right? What kind of tests should I add?

### DIFF
--- a/src/parameterized/gamma.jl
+++ b/src/parameterized/gamma.jl
@@ -1,0 +1,27 @@
+# Gamma distribution
+
+export Gamma
+
+@parameterized Gamma(α,β) ≪ Lebesgue(ℝ₊)
+
+@kwstruct Gamma(α, β)
+
+@kwalias Gamma [
+    a     => α
+    alpha => α
+    b     => β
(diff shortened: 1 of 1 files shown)


cscherrer:
That was quick! I think my biggest concern here is making sure the parameterization is clear. I've mostly defaulted to what Distributions uses, to make it easy for people coming from that. I think here we could have one parameterization in terms of shape and scale, and another for shape and rate. Then we just need to check around (Distributions, Wikipedia, Stan, etc) to be sure the parameter names we choose are reasonable common.

As for tests, I think we should have a standardized one for anything with a `distproxy`, something like
```julia
function distributions_test(d)
    b = rootmeasure(d)
    x = rand(d)
    @test logdensity(d, b, x) ≈ logdensity(distproxy(d), x)
end
```

Here `rootmeasure` just applies `basemeasure` until it hits a fixpoint, so in most parameterized measure it's `Lebesgue(ℝ)`.

gdalle:
Replicated in #147

PR #147, opened 2021-08-20
Euclidean space and multidimensional Lebesgue

gdalle:
Second attempt at this PR, I had a problem with my fork so I started again


src/parameterized/gamma.jl
```diff
@@ -0,0 +1,27 @@
+# Gamma distribution
+
```
cscherrer, line 2: I was just thinking it would be helpful to add some references to the tops of the files, something like
```suggestion
# REFERENCES
# https://mc-stan.org/docs/2_27/functions-reference/gamma-distribution.html
# https://en.wikipedia.org/wiki/Gamma_distribution
# https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.Gamma
```

src/parameterized/gamma.jl
```diff
@@ -0,0 +1,27 @@
...
+export Gamma
+
+@parameterized Gamma(α,β) ≪ Lebesgue(ℝ₊)
```
cscherrer, line 5: Do you see a reason not to use the Distributions parameterization for the default? We can have both of course

src/parameterized/gamma.jl
```diff
@@ -0,0 +1,27 @@
...
+    beta  => β
+]
+
+TV.as(::Gamma) = asℝ₊
+
+function logdensity(d::Gamma{(:α, :β)}, x)
+    return d.α * log(d.β) + (d.α - 1) * log(x) - d.β * x - loggamma(d.α)
+end
+
+Base.rand(rng::AbstractRNG, T::Type, μ::Beta) = rand(rng, Dists.Gamma(μ.α, 1/μ.β))
```
cscherrer, line 22: I think this isn't needed if we already have `distproxy` (but let's check that this is the case)

src/primitives/lebesgue.jl
```diff
@@ -13,15 +13,19 @@ end
 
 Lebesgue(X) = Lebesgue{X}()
 
+Lebesgue(es::EuclideanSpace) = Lebesgue(EuclideanSpace{StaticInt{dimension(es)}})
+
```
cscherrer, line 17: Let's add a `show` method for this, and maybe a method so `ℝ ^ (n::Int)` returns a `EuclideanSpace`

src/primitives/lebesgue.jl
```diff
@@ -13,15 +13,19 @@ end
...
 sampletype(::Lebesgue{ℝ}) = Float64
 sampletype(::Lebesgue{ℝ₊}) = Float64
 sampletype(::Lebesgue{𝕀}) = Float64
+sampletype(::Lebesgue{EuclideanSpace}) = Vector{Float64}
 
 testvalue(::Lebesgue{ℝ}) = 0.0
 testvalue(::Lebesgue{𝕀}) = 0.5
 testvalue(::Lebesgue{ℝ₊}) = 1.0
 testvalue(::Lebesgue{<:Real}) = 0.0
+testvalue(::Lebesgue{EuclideanSpace{StaticInt{D}}}) where {D} = zeros(Float64, D)
```
cscherrer, line 28: Note to self: Check this use of `StaticInt`

cscherrer (commented):
Thanks @gdalle ! I have a few comments, and some things I need to check before we merge

cscherrer:
Also, we need to add 
```julia
Static = "0.3"
```

near this line:
https://github.com/cscherrer/MeasureTheory.jl/blob/9e59beb75d36b909a1a99ab942117523b0ad4800/Project.toml#L45

src/primitives/lebesgue.jl
cscherrer: Added in 
https://github.com/cscherrer/MeasureTheory.jl/pull/147/commits/a11d9a6d54b2fa891ed4c0df159c2e764b30f813

cscherrer: Added in
https://github.com/cscherrer/MeasureTheory.jl/pull/147/commits/a11d9a6d54b2fa891ed4c0df159c2e764b30f813

src/parameterized/gamma.jl
cscherrer: Confirmed, let's drop this line

gdalle:
Sorry, I mistakenly put both the gamma dist and the euclidean space in the PR. Still new at open source :innocent:

src/parameterized/gamma.jl
gdalle: I spontaneously used the one I was familiar with, but you're right, for consistency sticking to Distributions is better style

cscherrer:
> Sorry, I mistakenly put both the gamma dist and the euclidean space in the PR. Still new at open source innocent

Haha yeah I saw that. It's ok, we don't need to be to picky, especially when the changes are mostly localized like this. :)

src/parameterized/gamma.jl
cscherrer: Ok let's have the Distributions form as the default, and also add the alpha,beta parameterization

Issue #145, opened 2021-08-20
Categorical and Gamma measures


gdalle:
Hey, it's me again :wave: 
In order to replace the `Distributions.jl` dependency in `PointProcesses.jl` with `MeasureTheory.jl`, I would need local implementations of the Gamma and Categorical distributions. Is that on the wish list somehow?
I'm not very confident with the macros defining `ParametrizedMeasure`s, so I could give it a shot in a PR but I'm not sure it would land on target.

cscherrer:
Yes! If you'd like to take a crack at it I can help if you get stuck, and of course review the PR. It would be really helpful to better understand the sticking points for this sort of thing.

I think I'd start by having a look at some of the existing parameterized measures to get a feel for it. I'm happy to help if something's not clear :)

PR #148, opened 2021-08-23
Categorical distribution

gdalle:
Partially solves #145. I did not add alternative (log-p) parametrizations yet.
I think this one is a typical example of the limits of our approach, which was to default on Distributions.jl for sampling and other things whenever possible. Indeed, their implementation of `Categorical` relies on `DiscreteNonParametric`, which is a nightmare in terms of efficiency (see [here](https://github.com/JuliaStats/Distributions.jl/blob/2856bc3e106d751a02e3fe9cccb046916a52955e/src/univariate/discrete/discretenonparametric.jl#L30) for the worst part). 
In my opinion, this boils down to their library being "safe" whereas we over at MeasureTheory.jl choose to trust that the user will avoid nonsensical inputs.

### DIFF
--- a/src/parameterized/categorical.jl
+++ b/src/parameterized/categorical.jl
@@ -1,0 +1,27 @@
+# Categorical distribution
+
+# REFERENCES
+# https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.Categorical
+# https://juliastats.org/Distributions.jl/stable/univariate/#Distributions.DiscreteNonParametric
+
+export Categorical
+
+@parameterized Categorical(p) ≪ CountingMeasure(ℤ[0:∞])
+
+ncategories(d::Categorical) = length(d.p)
+
+(d::Categorical ≪ ::CountingMeasure{IntegerRange{a,b}}) where {a,b} = a ≤ 1 && b ≥ ncategories(d)
+
+(::CountingMeasure{IntegerRange{a,b}} ≪ ::Categorical) where {a,b} = a ≥ 1 && b ≤ ncategories(d)
+
+###############################################################################
+@kwstruct Categorical(p)
+
+logdensity(d::Categorical{(:p)}, y) = log(d.p[y])
+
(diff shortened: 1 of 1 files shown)


cscherrer:
Thanks @gdalle for pushing on this, it's great to have it moving forward. Yes, I think we can get things much more efficient, in a few ways. First, in the `logp` parameterization, ...
1. the `logdensity` will just be `getindex`
2. we can use the Gumbel-max trick for sampling
3. We can avoid normalization, leaving it to depend on `logsumexp(logp)`

Then, in the case where the user passes something already normalized, it will turn out independent of the measure.

Clearly we also need the unlogged parameterization, and I think we can do much better in that case as well.

It's very strange to me that they `sortperm(vs)`. If anything, I would expect sorting the probabilities in reverse order, so when you walk down them to sample you can stop earlier. Anyway, I think we can avoid this, or when we do have to sort we can lean on PermutedArrays:
https://github.com/cscherrer/PermutedArrays.jl

I have a whole plan for using this to build a better DiscreteNonparametric based around a PriorityQueue, but implemented more efficiently than that. It will use Dictionaries.jl. That's a whole separate issue, so we can discuss separately if you like :)

cscherrer:
Also I should maybe point out that the whole `distproxy` thing isn't core to our approach. It's a stand-in, and when it works well we can keep it. But the great thing is, we can break away as needed to get better flexibility and/or performance

gdalle:
I think the `sortperm` is useful for quick `cdf` computations, because then you only need a `searchsortedfirst(vs, x)` to know at which step of the discrete distribution you are.

jeremiahpslewis:
Hey! Is this PR blocked or is there something manageable which still needs to be tackled?

gdalle:
Honestly I haven't touched this in years so I better let @cscherrer answer

irisallevi:
Hi! Any news on this?

gdalle:
Not on my end, my work no longer requires MeasureTheory so I don't think I'll pursue this further

gdalle:
Feel free to take it over though

PR #149, opened 2021-08-24
fix distproxy on exotic Normal parametrizations

jw3126:


jw3126:
Thanks, @cscherrer! So my motivation was that I actually wanted to use the `Normal(;μ,τ)` parametrization. Should I open a PR to MeasureBase.jl that adds methods for `AffineTransform{(:μ,:τ)}`? Or do you want to cut down the number of supported parametrizations for now?

cscherrer:
Hmm, I'm not sure yet. I've never thought about that parameterization outside the case of Normal distributions. If it makes sense in this case we should consider adding it, otherwise maybe just keep tau specific to normals.

@sethaxen or @mschauer , what do you think?

cscherrer:
I don't want to hold up your work with this, so let's go ahead with the tau parameterization as a special case for now. We might need to refactor this later. I think we'll just need a few small changes; I'll start a review.

Thanks again for working on this :)

mschauer:
In my naïveté,
```
function get_std(nt::NamedTuple)
     σ = get(nt, :σ, nothing)
     logσ = get(nt, :logσ, nothing)
     σ² = get(nt, :σ², nothing)
     τ = get(nt, :τ, nothing)
     if σ !== nothing
         return σ
     elseif logσ !== nothing
         return exp(logσ)
     elseif σ² !== nothing
         return √(σ²)
     elseif τ !== nothing
         return √(inv(τ))
     else
         return 1
     end
 end
 ```
feels like a human should not have to write that table

**Comment by jw3126** [comment]
I am closing this. I started merging the current master into it, but it does not play so well with the new Affine mechanism. I think the extra complexity added, may not be worth the benefit.

**Comment by cscherrer** [comment]
@jw3126 are there problems you had planned to address with this (or in general) that are still lingering?

**Comment by jw3126** [comment]
@cscherrer so my use case was just that a conjugated prior calculation was cleaner using precision (tau) rather than variance. But it's not a big deal to do it with any other parametrization either. 
What I felt was a little hairy with this PR was to implement `distproxy`. I think what needs to be done is something like
```julia
distproxy(::Affine) = Distributions.LocatationScale(...)
distproxy(::Affine{...Normal...}) = Distributions.Normal(...)
```
For this, you need to touch the internals of Affine a bit, but I thought Affine is maybe not yet in a stable form and this PR would add friction to further iterations on Affine.
Or maybe doing this PR does help shaping the design of Affine? In that case, I can continue.

Alternatively on could allow tau in Affine directly, as you remarked somewhere above. Just not sure if tau is standard for arbitrary location-scale families.

PR #156, opened 2021-09-02
fix half-measures

cscherrer:
TODO
- [x] Add tests

### DIFF
--- a/src/parameterized/studentt.jl
+++ b/src/parameterized/studentt.jl
@@ -1,7 +1,7 @@
 
 # StudentT distribution
 
-export StudentT
+export StudentT, HalfStudentT
 
 @parameterized StudentT(ν) ≪ Lebesgue(ℝ)
 
--- a/src/parameterized/studentt.jl
+++ b/src/parameterized/studentt.jl
@@ -6,12 +6,15 @@ export StudentT, HalfStudentT
 @parameterized StudentT(ν) ≪ Lebesgue(ℝ)
 
 @kwstruct StudentT(ν)
+@kwstruct StudentT(ν,μ)
+@kwstruct StudentT(ν,σ)
+@kwstruct StudentT(ν,μ,σ)
 
-StudentT(nt::NamedTuple{(:ν,:μ,:σ)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:μ,:ω)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:σ,)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:ω,)}) = Affine(nt, StudentT())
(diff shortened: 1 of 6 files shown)


cscherrer:
@sethaxen I think it's fixed, but the fix requires MeasureBase v0.3.7, which is pending registration. I assume we'll need to re-run the tests once that's ready.

src/parameterized/studentt.jl
```diff
@@ -1,17 +1,20 @@
...
+@kwstruct StudentT(ν,μ)
+@kwstruct StudentT(ν,σ)
+@kwstruct StudentT(ν,μ,σ)
 
-StudentT(nt::NamedTuple{(:ν,:μ,:σ)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:μ,:ω)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:σ,)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:ω,)}) = Affine(nt, StudentT())
-StudentT(nt::NamedTuple{(:ν,:μ,)}) = Affine(nt, StudentT())
+StudentT(nt::NamedTuple{(:ν,:μ,:σ)}) = Affine(NamedTuple{(:μ,:σ)}(nt), StudentT(ν=nt.ν))
```
sethaxen, line 13: I know this isn't part of this PR, but overloading constructors to construct something else can cause problems, the simplest being that something like `StudentT(0, 1, 30) isa StudentT` is `false`.

test/runtests.jl
```diff
@@ -396,3 +396,33 @@ end
...
+    end
+
+    @testset "HalfCauchy" begin
+        d = Cauchy(σ=3)
+        h = HalfCauchy(3)
+        x = rand(h)
+        @test density(h, Lebesgue(ℝ), x) ≈ 2 * density(d, Lebesgue(ℝ), x)
+    end
+
+    @testset "HalfLaplace" begin
```
sethaxen, line 415: Isn't the half-Laplace just the Exponential distribution?

sethaxen (approved):
Pending tests passing, this PR LGTM. I did have some extra-PR questions.

test/runtests.jl
cscherrer: Oh weird, hadn't ever thought of that. That sounds right, if it is we can just drop `HalfLaplace`

src/parameterized/studentt.jl
cscherrer: Yeah, I agree that's a little weird. But the upside is pretty huge. This approach allows us to fine-tune code for `Affine`, and have that same code work for multivariate-anything. So implementing MvNormal gives us MvStudentT etc for free.

To me it seems worth it, but if you disagree we should talk through it

cscherrer: Made this an issue:
https://github.com/cscherrer/MeasureTheory.jl/issues/157

I'll resolve this now, we can discuss there

PR #160, opened 2021-09-09
Update MvNormal

cscherrer:
This PR makes simplifications to `MvNormal`, based on the new `Affine` constructor in MeasureBase

### DIFF
--- a/src/parameterized/mvnormal.jl
+++ b/src/parameterized/mvnormal.jl
@@ -1,61 +1,29 @@
 
 # Multivariate Normal distribution
 
-using LinearAlgebra
 export MvNormal
-using Random
-import Base
 
-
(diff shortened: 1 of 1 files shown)


cscherrer:
I think it's good! This test passes:
```julia
@testset "MvNormal" begin
    Q,R = qr(randn(4,2))
    # Make sure `R` has a positive diagonal
    D = Diagonal(sign.(diag(R)))
    Q = Matrix(Q) * D
    R = D * R

    z = randn(2)
    
    ℓ = logpdf(MvNormal((σ= R,)),z)
    @test ℓ ≈ Dists.logpdf(Dists.MvNormal(R*R'),z)
    @test ℓ ≈ logpdf(MvNormal((σ= Q*R,)),Q*z)
end
```

But I broke a bunch of `rand!` stuff along the way, so I need to fix that before merging

PR #169, opened 2021-11-05
Doc adjustments

jeremiahpslewis:


jeremiahpslewis:
@gdalle Here’s my go through of the affine transform doc page. I’m new to the material so I may have introduced factual errors in pursuit of simplifying / adjusting. I defer to your judgement as to what to keep / what to skip. :)

docs/src/affine.md
```diff
@@ -1,27 +1,29 @@
...
+It is a common misconception that affine transformations are "linear". Linearity requires ``f(ax + by) = a f(x) + b f(y)`` for scalars ``a`` and ``b``, which would only holds for the above affine transformation ``f`` if ``μ=0``.
 
 
 ## Cholesky-based parameterizations
 
-If the "un-transformed" `z` is a scalar, things are relatively simple. But it's important our approach handle the multivariate case as well.
+If the "un-transformed" `z` is univariate, things are relatively simple. But it's important our approach handle the multivariate case as well.
 
-In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but can be very awkward from a computational perspective.
+In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but less idea for efficient computation.
```
cscherrer, line 20: It seems like the point here is that `awkward` is vague. I agree with that, how about
```suggestion
In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but leads an ``O(n^3)`` [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), which can be a significant bottleneck.
```

docs/src/affine.md
```diff
@@ -1,27 +1,29 @@
...
 
 While MeasureTheory.jl includes (or will include) a parameterization using `Σ`, we prefer to work in terms of its Cholesky decomposition ``σ``.
 
-Using "``σ``" for this may seem strange at first, so we should explain the notation. Let ``σ`` be a lower-triangular matrix satisfying
+The relationship between the computationally efficient "``σ``" and more familiar parameteriation `Σ` can be seen as follows:
+
+Let ``σ`` be a lower-triangular matrix satisfying
```
cscherrer, line 26: I think I see where you're going with this, and I like it. But maybe we can keep the "relationship" perspective and also the conciseness with 
```suggestion
To see the relationship between our ``σ`` parameterization and the likely more familiar  ``Σ`` parameterization,  let ``σ`` be a lower-triangular matrix satisfying
```

docs/src/affine.md
```diff
@@ -1,27 +1,33 @@
...
 
 ```math
 f(z) = σ z + μ
 ```
 
-Here are below, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure like `Normal()` with no location or scale parameters.
+Starting with the above definition, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure which has neither a location nor a scale parameter, for example `Normal()`.
 
-Affine transforms are often incorrectly referred to as "linear". Linearity requires ``f(ax + by) = a f(x) + b f(y)`` for scalars ``a`` and ``b``, which only holds for the above ``f`` if ``μ=0``.
+Affine transforms are often mistakenly referred to as "linear". For a function `f` to be linear requires
```
jeremiahpslewis, line 13:
```suggestion
Affine transformations are often mistakenly referred to as "linear". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation). For a function `f` to be linear requires 
```

docs/src/affine.md
```diff
@@ -1,27 +1,33 @@
...
 ```
 
-Here are below, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure like `Normal()` with no location or scale parameters.
+Starting with the above definition, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure which has neither a location nor a scale parameter, for example `Normal()`.
 
-Affine transforms are often incorrectly referred to as "linear". Linearity requires ``f(ax + by) = a f(x) + b f(y)`` for scalars ``a`` and ``b``, which only holds for the above ``f`` if ``μ=0``.
+Affine transformations are often mistakenly referred to as "linear". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation). For a function `f` to be linear requires 
+``f(ax + by) == a f(x) + b f(y)``
+for scalars ``a`` and ``b``. For an affine function
+``f(z) = σ * z + μ``,
```
jeremiahpslewis, line 16:
```suggestion
``f(z) = σ * z + μ``, where the linear map is defined as `σ` and the translation defined as `μ`,
```

docs/src/affine.md
```diff
@@ -1,27 +1,33 @@
...
+linearity holds only if ``μ = 0``.
```
jeremiahpslewis, line 17:
```suggestion
linearity holds only if the translation component ``μ`` is equal to zero.
```

docs/src/affine.md
```diff
@@ -1,27 +1,33 @@
...
+linearity holds only if the translation component ``μ`` is equal to zero.
 
 
 ## Cholesky-based parameterizations
 
-If the "un-transformed" `z` is a scalar, things are relatively simple. But it's important our approach handle the multivariate case as well.
+If the "un-transformed" `z` is univariate, things are relatively simple. But it's important our approach handle the multivariate case as well.
 
-In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but can be very awkward from a computational perspective.
+In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but leads an ``O(n^3)`` [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), which can be a significant bottleneck.
```
jeremiahpslewis, line 24:
```suggestion
In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but leads an ``O(n^3)`` [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), which becomes a significant bottleneck to compute as ``n`` gets large.
```

jeremiahpslewis:
@cscherrer Thanks for the review, I went ahead and added your suggestions and made a few further tweaks based on the direction of your feedback. Looking forward to hearing from @gdalle :)

docs/src/affine.md
```diff
@@ -1,27 +1,31 @@
...
 
 ```math
 f(z) = σ z + μ
 ```
 
-Here are below, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure like `Normal()` with no location or scale parameters.
+Starting with the above definition, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure which has neither a location nor a scale parameter, for example `Normal()`.
 
-Affine transforms are often incorrectly referred to as "linear". Linearity requires ``f(ax + by) = a f(x) + b f(y)`` for scalars ``a`` and ``b``, which only holds for the above ``f`` if ``μ=0``.
+Affine transformations are often mistakenly referred to as "linear". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation). For a function `f` to be linear requires
```
mschauer, line 13:
```suggestion
Affine transformations are often ambiguously referred as "linear transformations". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation) in the stricter algebraic sense: For a function `f` to be linear requires 
```

jeremiahpslewis: Thanks!

cscherrer: Looks like you may have accepted just as Moritz was suggesting an alternative. I think I like his suggestion better than mine, though I'd keep the "as" (maybe a US vs European difference?)

docs/src/affine.md
```diff
@@ -1,27 +1,31 @@
...
 
 ```math
 f(z) = σ z + μ
 ```
 
-Here are below, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure like `Normal()` with no location or scale parameters.
+Starting with the above definition, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure which has neither a location nor a scale parameter, for example `Normal()`.
 
-Affine transforms are often incorrectly referred to as "linear". Linearity requires ``f(ax + by) = a f(x) + b f(y)`` for scalars ``a`` and ``b``, which only holds for the above ``f`` if ``μ=0``.
+Affine transformations are often ambiguously referred as "linear transformations". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation) in the stricter algebraic sense: For a function `f` to be linear requires
```
gdalle, line 13: Is this paragraph really useful? I would assume most of our readers know the difference between linear and affine

docs/src/affine.md
```diff
@@ -1,27 +1,31 @@
...
+linearity holds only if the translation component ``μ`` is equal to zero.
 
 
 ## Cholesky-based parameterizations
 
-If the "un-transformed" `z` is a scalar, things are relatively simple. But it's important our approach handle the multivariate case as well.
+If the "un-transformed" `z` is univariate, things are relatively simple. But it's important our approach handle the multivariate case as well.
 
-In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but can be very awkward from a computational perspective.
+In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but leads an ``O(n^3)`` [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), which becomes a significant bottleneck to compute as ``n`` gets large.
```
gdalle, line 24: leads to

docs/src/affine.md
```diff
@@ -1,27 +1,31 @@
...
 
 While MeasureTheory.jl includes (or will include) a parameterization using `Σ`, we prefer to work in terms of its Cholesky decomposition ``σ``.
 
-Using "``σ``" for this may seem strange at first, so we should explain the notation. Let ``σ`` be a lower-triangular matrix satisfying
+To see the relationship between our ``σ`` parameterization and the likely more familiar  ``Σ`` parameterization,  let ``σ`` be a lower-triangular matrix satisfying
 
 ```math
 σ σᵗ = Σ
```
gdalle, line 31: I don't like this notation for the transpose, I very much prefer $\sigma^\top$, is it the Frenchman in me?

docs/src/affine.md
```diff
@@ -33,23 +37,23 @@ Then given a (multivariate) standard normal ``z``, the covariance matrix of ``σ
...
 
 ```math
 x = σ z + μ
 ```
 
-On the other hand, the log-density computation is not quite so simple. Starting with an ``x``, we need to find ``z`` using
+The log-density transformation of a `Normal` with parameters μ, σ does not follow as directly. Starting with an ``x``, we need to find ``z`` using
 
 ```math
 z = σ⁻¹ (x - μ)
```
gdalle, line 59: I think there is a problem with the exponent in the maths formula, why not use $\sigma^{-1}$?

gdalle (commented):
I mainly spotted minor typos, but a more important comment would be the name `Affine`: why not `AffineMeasure`?

docs/src/affine.md
mschauer: Yes, but we should say what *affine* is anyway. But maybe just a brief definition and a link.

*An [affine function](https://en.wikipedia.org/wiki/Affine_transformation) `f(x) = A*x + b`  is just a linear map `A` translated by a constant `b` with constant `b` and linear map/operator `A` .

cscherrer:
> I mainly spotted minor typos, but a more important comment would be the name `Affine`: why not `AffineMeasure`?

We'll definitely have some things with `Measure` explicitly in the name (`CountingMeasure`), and others without (`Normal`). I'm not sure I've been entirely consistent on this, but generally my approach has been to prefer the shorter name unless it's confusing. We should talk more about this if you or others have other suggestions.

docs/src/affine.md
cscherrer: One benefit of having it as unicode when possible is that it can render in the REPL. But we should check that the docs also render well (do they not?)

docs/src/affine.md
cscherrer: I think those knowing the difference are limited to a small neighborhood around "math PhD programs". We need to be very clear that there's a difference (otherwise many will think we're just using `Affine` to sound fancy). But the original might be too pedantic about it. Having something more brief is fine, and maybe preferable.

docs/src/affine.md
cscherrer: I always found "⊤ is $\top$, and that starts with a T, and that stands for transpose" kind of funny. But I'm open to changing it to anything widely used ($\top$ included), as long as we're consistent

gdalle: Alternately, $\sigma^T$ would be fine. Anything but the small t works for me ^^

docs/src/affine.md
gdalle: I think they're ok

docs/src/affine.md
gdalle:
```suggestion
In the literature, it's common for a multivariate normal distribution to be parameterized by a mean `μ` and covariance matrix `Σ`. This is mathematically convenient, but leads to an ``O(n^3)`` [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), which becomes a significant bottleneck to compute as ``n`` gets large.
```

docs/src/affine.md
cscherrer: This is a simple search and replace (superscripts make it convenient in this case). But maybe a separate PR would be easier? Otherwise it could be easy to miss some cases as we're discussing other changes

docs/src/affine.md
jeremiahpslewis: Think it’s now incorporated. :)

docs/src/affine.md
jeremiahpslewis: Noted here: https://github.com/cscherrer/MeasureTheory.jl/issues/173

gdalle (approved):

gdalle:
Do we really have to issue a new release for such small, docs-only changes @cscherrer?

cscherrer:
Not critical, but also not much reason not to have a minor release. They're free and pretty easy 🙂

cscherrer:
Not critical, but also not much reason not to have a minor release. They're free and pretty easy 🙂

The one reason I see it could be nice is so people are more likely to see the new version of the docs

PR #171, opened 2021-11-08
Remove duplicate method

keorn:
Was getting warnings about method overwrite given the definition right below the deleted one.


cscherrer (approved):
Good catch, thanks @keorn :)

PR #182, opened 2022-01-16
Add a `Uniform(a, b)` measure

jarredbarber:
This PR adds a parameterized uniform measure on an interval `(a, b)`.

### DIFF
--- a/src/parameterized/uniform.jl
+++ b/src/parameterized/uniform.jl
@@ -64,6 +64,6 @@ end
 
 distproxy(d::Uniform{(:shift, :scale)}) = Dists.Uniform(d.shift, d.shift + d.scale)
 
-TV.as(d::Uniform{(:shift, :scale), T}) where T = TV.ScaledShiftedLogistic(Real, d.scale, d.shift)
+TV.as(d::Uniform{(:shift, :scale)}) = TV.ScaledShiftedLogistic(d.scale, d.shift)
 
 Base.rand(rng::AbstractRNG, T::Type, μ::Uniform{(:shift, :scale)}) = rand(rng, T)*μ.scale + μ.shift
--- a/src/parameterized/uniform.jl
+++ b/src/parameterized/uniform.jl
@@ -28,3 +28,40 @@ Base.rand(rng::AbstractRNG, T::Type, μ::Uniform{()}) = rand(rng, T)
 
 ###############################################################################
 # Uniform
+#
+# The parameterization was chosen to enforce the constraint that the parameters
+# define an interval. Since `asparams` operates on parameters independently,
+# we parameterize with the left point of the interval (lower bound) and the width
+# of the interval, which is parameterized as positive real number (`asℝ₊`).
+
+@kwstruct Uniform(left, width)
+
+function basemeasure(d::Uniform{(:left, :width)})
+    # Lebesgue measure on (left, width+left)
+    lb = d.left
+    ub = d.left + d.width
+    inbounds(x) = lb < x < ub
+    constℓ = 0.0
+    varℓ() = 0.0
+    base = Lebesgue(ℝ)
+    FactoredBase(inbounds, constℓ, varℓ, base)
+end
+
+asparams(::Type{<:Uniform{(:left, :width)}}, ::Val{:left}) = asℝ
+asparams(::Type{<:Uniform{(:left, :width)}}, ::Val{:width}) = asℝ₊
+
+"""
+The uniform probability measure on the interval (a, b).
+"""
+function Uniform(a, b)
+    @assert(a < b)
+    Uniform(left=a, width=b-a)
+end
+
+logdensity(d::Uniform{(:left, :width)}, x) = -log(d.width)
+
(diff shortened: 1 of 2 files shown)


jarredbarber:
It might be cleaner to just add a function `Uniform(a, b)` that constructs an `Affine`; I did it this way because (1) I wasn't sure how that would interact with `inbounds` and (2) I was following the `Normal` distribution as an example and was trying to stick to how that worked.

src/parameterized/uniform.jl
```diff
@@ -27,4 +27,43 @@ TV.as(::Uniform{()}) = as𝕀
...
+"""
+The uniform probability measure on the interval (a, b).
+"""
+function Uniform(a, b)
+    @assert(a < b)
+    Uniform(shift=a, scale=b-a)
+end
+
+logdensity(d::Uniform{(:shift, :scale)}, x) = begin
+    @assert (d.scale > 0.0)
```
chriselrod, line 61: Perhaps these should be checked in inner constructor to avoid the check on calls to `logdensity`.

I prefer invalid values such as `-Inf` or `NaN` over errors.
Density functions are used in many iterative algorithms that can on occasion produce invalid inputs.
`-Inf` or `NaN` will cause these algorithms to reject the offending parameter values and move on.
Errors will cause termination, or require expensive `try`/`catch` and then extra effort to avoid type instabilities.

chriselrod: I haven't discussed this with @cscherrer , so I don't know his take, but it is my number 1 gripe with Distributions.jl.
I have to add code like `isnan(x) ? 0.0 #= results in -Inf =# :  abs(x)` for values that must be positive all over the place to work around it, or opt in via keyword arguments.
Neither of which is something I can rely on end users doing.
Meaning it may be worth eventually abandoning Distributions.jl, or trying to add passes that make the necessary adjustments.

All a bunch of extra work and complexity.

cscherrer: Thank you both! Yes, I agree it's good to avoid try/catch. In this case, I think it's not really a problem to have a negative scale, since that's really just a linear coefficient in an affine transform. 

BTW, things in the works are in the `dev2` branch here and in MeasureBase. One thing that seems worth discussing here is the idea of a `proxy`.

For a while `Normal` was its own thing. Lots of parameterizations, defined independently. But that's kind of annoying, since scale/location families come up all over the place, and it's weird to implement them so many times.

That led to development of `Affine`. But then it's equally weird to ask for a `Normal(2,3)` and get an `Affine` as a result. Maybe we want to send the result into the function, so we'll need dispatch. It's good to avoid having to dig too deeply  into the type, when we have a chance.

So now (in `dev2`), `Normal(2,3)` builds a `Normal`, but `basemeasure` or `logdensity_def` just forward that `Normal` to its `proxy`, which is an `Affine`. Like this:
```julia
julia> d = Normal(2,3)
Normal(μ = 2, σ = 3)

julia> MeasureBase.proxy(d)
Affine((μ = 2, σ = 3), Normal{(), Tuple{}}(()))
```
[That should be `Affine((μ = 2, σ = 3), Normal())`, looks like a `show` bug]

Lately I've been working on composability and performance for exponential families, so some of this will change again. But that's not so relevant for `Uniform` unless the bounds are fixed.

Also, I should mention that new releases are AGPL, so I can offer dual licensing for commercial users. This is in an attempt to make it more likely I'll be able to keep doing this work. For that purpose I think it's the right call, but unfortunately it does mean I'll need to get something set up for people to contribute, in order to allow those contributions to be relicensed commercially. I'm still learning about all of this, but I'd be happy to discuss if you have any feedback or ideas.

cscherrer:
Hey @jarredbarber, I have a few updates...

First, I'm punting on the AGPL thing, going back to MIT. Sales would be a ways off, and it hit me that I really want to avoid being in such a sales role anyway. And it was feeling kind of isolating, which isn't great either. 

Also, tracking the support as part of the log-density wasn't working well, so I've pulled that out. So now `insupport` is separate from `logdensity_def`, and `logdensityof` puts everything together.

So having another look at this today, I found it's now pretty easy to implement this. We really just need

```julia
@kwstruct Uniform(a,b)
Uniform(a,b) = Uniform((a=a,b=b))

insupport(d::Uniform{(:a,:b)}, x) = d.a ≤ x ≤ d.b

proxy(d::Uniform{(:a,:b)}) = affine((μ=d.a, σ=d.b - d.a), Uniform())

@useproxy Uniform{(:a,:b)}
Base.rand(rng::Random.AbstractRNG, ::Type{T}, μ::Uniform) where {T} = rand(rng, T, proxy(μ))
```

Here `@kwstruct` is from `KeywordCalls.jl`, and the second line makes it so we can use this with unnamed arguments. The `@useproxy` line tells MeasureTheory to use the proxy for calls to `logdensity_def` and `basemeasure`. This doesn't take care of `rand`, so I have that line separately.

This will make it into the `dev` branch soon.

cscherrer:
Added support for `Uniform(a,b)` in version 0.15, so I'll close this now. Thanks @jarredbarber

PR #201, opened 2022-05-20
Cs edits

mschauer:
See comments in the diff (I just made a formal PR to get a diff to comment on)


paper/paper.tex
```diff
@@ -0,0 +1,946 @@
...
+Our primary interest, and the goal of \verb|MeasureTheory|, is to offer support for \emph{applied probabilistic modeling}. 
+The package name is owed to the fact, that the word measure alone is too vague.
+
+Given a space $X$ of possible outcomes and a set of subsets of $X$ called \href{https://en.wikipedia.org/wiki/Event_(probability_theory)}{\emph{events}}, a probability distribution assigns each event a non-negative quantity, called the \emph{probability} or \emph{probability mass}.
+Likewise, a \emph{measure} assigns each \emph{measurable set} a non-negative quantity, also called a ``mass''.
+
+The space $X$ could be the space $\{1,2,3,4,5,6\}$ suitable to model a six-sided die, or $X$ might be the 3-dimensional Euclidean space $\RR^3$ suitable to model ``volume'', to give two examples.
+
+One further such space that deserves mention is traditionally denoted $\Omega$. This is an abstract space connected to the real-world notion of \emph{probability}, which we denote as $\operatorname{Prob}$.
+Computationally, $\Omega$ can be considered to be the set of possible states of a random number generator \verb|rng::AbstractRNG|, which is the source of computational randomness.
```
mschauer, line 196:
```suggestion
Computationally, $\Omega$ can be considered to be the set of possible initial states of a random number generator \verb|rng::AbstractRNG|, which is the source of computational randomness.
```

paper/paper.tex
```diff
@@ -0,0 +1,946 @@
...
+
+A key observation is that having defined probability mass for a number of relatively simple sets, for example on all intervals,
+Kolmogorov's axioms give a recipe to compute the probability mass of more complicated sets, and with those, even more in new applications of the axioms, etc.
+One says the $\sigma$-algebra is \emph{generated} by these sets.
+Indeed, \verb|Distributions.jl| does not give means to compute probability for arbitrary subsets of floating point numbers,
+but gives a way of computing probability mass just of intervals of the form $(-\infty, x]$ through the function \verb|cdf|.
+
+For spaces that are ``small enough'', a meaningful probability can be assigned to each subset of the space. Unfortunately, this is impossible for any space containing a continuum such as the real line. Just representing a single arbitrary subset of the \verb|Float64|-range, the computational abstraction of the real line, would require staggering 2\,306 Petabytes of memory. This is not practical.
```
mschauer, line 294:
```suggestion
For spaces that are ``small enough'', a meaningful probability can be assigned to each subset of the space. 
But even in moderately sized spaces this can become infeasible: Just representing a single arbitrary subset of the \verb|Float64|-range, the computational abstraction of the real line, would require staggering 2\,306 Petabytes of memory. This is not practical. And for any space containing a continuum such as the real line it becomes not impractical but mathematically impossible to assign a meaningful probability to all subsets. 
```

paper/paper.tex
```diff
@@ -0,0 +1,946 @@
...
+The practical importance of $\sigma$-algebras here lies therein that they also encode available information. Observe that each event $A$ corresponds to a question, with a probabilistic answer. 
+ For example, an event $A = \{\bX \in  [c, d]\}$ corresponds to the question whether the random variable $\bX$ is in the interval $[c, d]$ and there is a probability $\operatorname{Prob}(A)$ that this will be the case. 
+ Assume we know whether some of such questions $A$, $B$ are answered (each with yes or no).
+ 
+ Then we also know that the answer to the complement $A^c$ (no or yes), and we always know that $\Omega$ has probability 1 and the corresponding question is answered by `yes'. Likewise, if both $A$ and $B$ can be answered given what we know, we also know the answer to $A \cup B$. 
+ 
+ As the reader might now suspect, such known events form a $\sigma$-algebra. In particular, for each random variable $\bX$ there is a $\sigma$-algebra associated with $\bX$ that encodes what else we know knowing $\bX$ and how we should assign (conditional) probabilities to other events knowing $\bX$.
```
mschauer, line 307:
```suggestion
 As the reader might now suspect, such systems of known events form a $\sigma$-algebra. In particular, for each random variable $\bX$ there is a $\sigma$-algebra associated with $\bX$ that encodes what else we know knowing $\bX$ and how we should assign (conditional) probabilities to other events knowing $\bX$.
```

paper/paper.tex
```diff
@@ -0,0 +1,946 @@
...
+
+This requires solving a linear system. Even with $\sigma$ being lower-triangular, this involves division operations and the allocation of a temporary vector for storage of $z$.
+
+In many cases, we prefer the density (or log-density, really) to be fast to evaluate. This leads us to a kind of dual approach to the above.
+
+\subsection*{Inverse parameterization}
+
+An alternative parameterization of a multivariate Gaussian is in terms of its \href{https://en.wikipedia.org/wiki/Precision_(statistics)}{\emph{precision matrix}}, $\Lambda = \VV[\bX]^{-1}$. Similarly to above, we'll write $\lambda$ for the lower Cholesky factor of $\Lambda$, so $\lambda \lambda^t = \Lambda$. The parameterization is then specified by
+\[
+\bZ = \lambda(\bX - \mu)\ .
```
mschauer, line 818:
```suggestion
\bZ = \lambda\cdot(\bX - \mu)\ .
```

paper/paper.tex
```diff
@@ -0,0 +1,946 @@
...
+The normalization term is present for any normal distribution, even those with a different mean or variance. So if we have a product of normals, for example in a regression problem, the normalization can be shared across them. For the simple example of $n$ iid standard \verb|Normal|s, this effectively rewrites the log-pdf
+\[
+    \sum_{j=1}^n \left(-\frac{1}{2}\log 2π -\frac{x_j^2}{2}\right)
+    \quad \text{as} \quad
+    -\frac{n}{2}\log 2π + \sum_{j=1}^n \left(-\frac{x_j^2}{2}\right)
+\]
+
+Further simplification is clearly possible; doing this in a composable way without sacrificing efficiency is the subject of ongoing work.
+
+\subsection*{The end-user perspective}
```
mschauer, line 417: Needs a warning that for inference it is important if parameters are in the base measure - dominated experiment and a way to query that

mschauer (commented):
Some comments and suggestions

paper/paper.tex
cscherrer: why \cdot? This is matrix-vector multiplication

paper/paper.tex
cscherrer: Can you give some more detail? Not sure I understand

paper/paper.tex
mschauer: I just thought it looks like a function application otherwise. For me \cdot is just multiplication, but perhaps we need to be more careful if we used it somewhere else?

paper/paper.tex
mschauer: If N(\mu, \sigma) has a base measure N(0, \sigma), this is good when estimating mu, but tricky when estimating $sigma$

paper/paper.tex
cscherrer: Hmm, I haven't thought of \cdot outside of vector-vector. In that case it involves an adjoint, so wouldn't it have an adjoint in general?

How about we add a `\<space>` and then a comment after that it's matrix-vector multiplication?

paper/paper.tex
cscherrer: Ah right, yes this comes up a lot. If we have two measures that depend on some variable and want to compare them, we need to use a base measure that's invariant wrt that variable. It's kind of like in physics how you want an inertial frame of reference.

paper/paper.tex
mschauer: just add \,

paper/paper.tex
mschauer: Just write that!

Issue #217, opened 2022-07-22
`logdensityof` fails for `For` with arrays


jfb-h:
When using the `For` constructor with multiple array inputs, like so:

```julia 
d = For(rand(3), rand(3)) do x, y
	Normal(x, y)
end
```
`d` is a `For{Any}(#9, ([0.553963, 0.567082, 0.450658], [0.197881, 0.0258519, 0.00209304]))`

and calling `logdensityof(d, rand(d))` throws this error:

```julia
MethodError: no method matching (::MeasureTheory.var"#13#14"{MeasureTheory.For{Any, var"#9#10"{Module}, Tuple{Vector{Float64}, Vector{Float64}}}})(::Float64, ::Float64)

Closest candidates are:

(::MeasureTheory.var"#13#14")(::Any) at C:\Users\Jakob\.julia\packages\MeasureTheory\MeOXc\src\combinators\for.jl:121

    (::MeasureTheory.var"#9#12"{MeasureTheory.For{Union{}, MeasureTheory.var"#13#14"{MeasureTheory.For{Any, var"#9#10"{Module}, Tuple{Vector{Float64}, Vector{Float64}}}}, Tuple{Vector{Float64}, Vector{Float64}}}, Vector{Float64}, MeasureTheory.var"#get_i#10"{MeasureTheory.For{Union{}, MeasureTheory.var"#13#14"{MeasureTheory.For{Any, var"#9#10"{Module}, Tuple{Vector{Float64}, Vector{Float64}}}}, Tuple{Vector{Float64}, Vector{Float64}}}}})(::Int64)@for.jl:76
    _mapreduce(::MeasureTheory.var"#9#12"{MeasureTheory.For{Union{}, MeasureTheory.var"#13#14"{MeasureTheory.For{Any, var"#9#10"{Module}, Tuple{Vector{Float64}, Vector{Float64}}}}, Tuple{Vector{Float64}, Vector{Float64}}}, Vector{Float64}, MeasureTheory.var"#get_i#10"{MeasureTheory.For{Union{}, MeasureTheory.var"#13#14"{MeasureTheory.For{Any, var"#9#10"{Module}, Tuple{Vector{Float64}, Vector{Float64}}}}, Tuple{Vector{Float64}, Vector{Float64}}}}}, ::typeof(Base.add_sum), ::IndexLinear, ::Base.OneTo{Int64})@reduce.jl:427
[...]

PR #218, opened 2022-07-22
fix splatting problem

cscherrer:
Fixes https://github.com/cscherrer/MeasureTheory.jl/issues/217


cscherrer:
Let's add the example from https://github.com/cscherrer/MeasureTheory.jl/issues/217 to tests before merging

PR #219, opened 2022-07-31
Fix cdf and univariate affine broadcasting

jeremiahpslewis:
Currently, the following code errors: 

```julia
cdf(MeasureTheory.Normal(37.5, 7.5), [10, 25])
```

Output:

```julia
MethodError: no method matching +(::Vector{Float64}, ::Float64)
For element-wise addition, use broadcasting with dot syntax: array .+ scalar
Closest candidates are:
+(::Any, ::Any, !Matched::Any, !Matched::Any...)
@ Base operators.jl:578
+(!Matched::T, ::T) where T<:Union{Float16, Float32, Float64}
@ Base float.jl:383
+(!Matched::Union{InitialValues.NonspecificInitialValue, InitialValues.SpecificInitialValue{typeof(+)}}, ::Any)
@ InitialValues ~/.julia/packages/InitialValues/OWP8V/src/InitialValues.jl:154
...
(::MeasureTheory.AffineTransform{(:μ, :λ), Tuple{Float64, Float64}})(::Vector{Int64})@affine.jl:61
cdf@affine.jl:340[inlined]
cdf(::MeasureTheory.Normal{(:μ, :σ), Tuple{Float64, Float64}}, ::Vector{Int64})@distproxy.jl:45
top-level scope@[Local: 1](http://localhost:1234/edit?id=1ffc8736-1004-11ed-3766-0df61498ed70#)[inlined]
```

### DIFF
--- a/src/combinators/affine.jl
+++ b/src/combinators/affine.jl
@@ -54,11 +54,19 @@ end
 
 Base.size(f::AffineTransform, n::Int) = @inbounds size(f)[n]
 
+# Univariate case with single `x`` observation
 (f::AffineTransform{(:μ,)})(x) = x + f.μ
 (f::AffineTransform{(:σ,)})(x) = f.σ * x
 (f::AffineTransform{(:λ,)})(x) = f.λ \ x
 (f::AffineTransform{(:μ, :σ)})(x) = f.σ * x + f.μ
 (f::AffineTransform{(:μ, :λ)})(x) = f.λ \ x + f.μ
+
+# Vector `x` observations broadcasts, adjusted for univariate and multivariate cases
+(f::AffineTransform{(:μ,)})(x::AbstractArray) = supportdim(params(f)) == () ? x .+ f.μ : x .+ (f.μ,)
+(f::AffineTransform{(:σ,)})(x::AbstractArray) = supportdim(params(f)) == () ? f.σ .* x : (f.σ,) .* x
+(f::AffineTransform{(:λ,)})(x::AbstractArray) = supportdim(params(f)) == () ? f.λ .\ x : (f.λ,) \ x
+(f::AffineTransform{(:μ, :σ)})(x::AbstractArray) = supportdim(params(f)) == () ? f.σ .* x .+ f.μ : (f.σ,) * x + (f.μ,)
+(f::AffineTransform{(:μ, :λ)})(x::AbstractArray) = supportdim(params(f)) == () ? f.λ .\ x .+ f.μ : (f.λ,) \ x + (f.μ,)
 
 rowsize(x) = ()
 rowsize(x::AbstractArray) = (size(x, 1),)
@@ -337,5 +345,5 @@ end
 end
 
 @inline function Distributions.cdf(d::Affine, x)
-    cdf(parent(d), inverse(d.f).(x))
+    cdf(parent(d), inverse(d.f)(x))
 end
--- a/src/combinators/affine.jl
+++ b/src/combinators/affine.jl
@@ -339,7 +339,3 @@ end
 @inline function Distributions.cdf(d::Affine, x)
     cdf(parent(d), inverse(d.f)(x))
 end
-
-# @inline function Distributions.cdf.(d::Affine, x::AbstractArray)
-#     cdf.((parent(d),), inverse(d.f)(x))
-# end
--- a/src/combinators/affine.jl
+++ b/src/combinators/affine.jl
@@ -348,6 +348,6 @@ end
     cdf(parent(d), inverse(d.f)(x))
 end
 
-@inline function Distributions.cdf(d::Affine, x::AbstractArray)
-    cdf.((parent(d),), inverse(d.f)(x))
-end
+# @inline function Distributions.cdf.(d::Affine, x::AbstractArray)
+#     cdf.((parent(d),), inverse(d.f)(x))
+# end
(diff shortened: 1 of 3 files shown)


jeremiahpslewis:
Note to self: this doesn't work for multivariate cases.

cscherrer:
Thanks for the PR! I haven't looked at the code yet, but...

> `cdf(MeasureTheory.Normal(37.5, 7.5), [10, 25])`

Wouldn't this need to be `cdf.(...)`?

jeremiahpslewis:
Fair question. I'm borrowing the API from Distributions.jl, as the issue came up in one of those cases where I had working Dist code and wanted to swap it over to MeasureTheory code. If you do the . broadcast in dist.jl you get a deprecation warning, so I wanted to keep aligned / in compliance with their standard.

cscherrer:
Thanks for the background. In general, matching Distributions is not a high priority, but in this particular case it seems complicated by the fact that we're adding methods to an existing function "owned" by Distributions.

I'd really rather avoid the Distributions approach of implicit broadcasting, but I'm not sure there's a clean way to do that here. @oschulz and @mschauer , what do you think?

cscherrer:
Hmm, what Distributions version are you using @jeremiahpslewis ?
```julia
julia> cdf.(Dists.Normal(), [2,3,4])
3-element Vector{Float64}:
 0.97725
 0.99865
 0.999968

(MeasureTheory) pkg> st Distributions
Project MeasureTheory v0.16.5
Status `~/git/MeasureTheory.jl/Project.toml`
  [31c24e10] Distributions v0.25.66
```

cscherrer:
It looks like the deprecation goes in the other direction. I don't see a warning (weird), but lowering your example gives
```julia
julia> @code_lowered cdf(Dists.Normal(37.5, 7.5), [10, 25])
CodeInfo(
1 ─      $(Expr(:meta, :noinline))
│   %2 = Core.Typeof
│   %3 = (%2)(Distributions.cdf)
│   %4 = Base.getproperty(%3, :name)
│   %5 = Base.getproperty(%4, :mt)
│   %6 = Base.getproperty(%5, :name)
│        Base.depwarn("`cdf(d::UnivariateDistribution, X::AbstractArray)` is deprecated, use `cdf.(d, X)` instead.", %6)
│   %8 = Base.broadcasted(Distributions.cdf, d, X)
│   %9 = Base.materialize(%8)
└──      return %9
)
```

In MeasureBase, I think we should add
```julia
Broadcast.broadcastable(m::AbstractMeasure) = Ref(m)
```

oschulz:
> I'd really rather avoid the Distributions approach of implicit broadcasting

Yes, please, *no* implicit broadcasting! It messes up dispatch, and Distributions.jl want's go get away from it too, I think.

jeremiahpslewis:
Ok, so incorporating the feedback, here’s a new take (see the tests). Right now the ``(dist,)`` syntax is required to broadcast over a distribution, but I guess that the MeasureBase PR @cscherrer created might remove this requirement.

test/runtests.jl
```diff
@@ -574,6 +574,14 @@ end
     f = AffineTransform((μ = 3,))
     @test f(inverse(f)(1)) == 1
     @test inverse(f)(f(1)) == 1
+
+    f = AffineTransform((μ = 3,))
+    @test f(inverse(f)(1)) == 1
+    @test inverse(f)(f(1)) == 1
```
cscherrer, line 580: Duplicated code?
```suggestion
```

src/combinators/affine.jl
```diff
@@ -54,12 +54,20 @@ end
...
 (f::AffineTransform{(:λ,)})(x) = f.λ \ x
 (f::AffineTransform{(:μ, :σ)})(x) = f.σ * x + f.μ
 (f::AffineTransform{(:μ, :λ)})(x) = f.λ \ x + f.μ
 
+# Broadcast over vector `x` for univariate distributions
+(f::AffineTransform{(:μ,)})(x::AbstractArray) = supportdim(params(f)) == () ? x .+ f.μ : x + f.μ
+(f::AffineTransform{(:σ,)})(x::AbstractArray) = supportdim(params(f)) == () ? f.σ .* x : f.σ * x
+(f::AffineTransform{(:λ,)})(x::AbstractArray) = supportdim(params(f)) == () ? f.λ .\ x : f.λ \ x
+(f::AffineTransform{(:μ, :σ)})(x::AbstractArray) = supportdim(params(f)) == () ? f.σ .* x .+ f.μ : f.σ * x + f.μ
+(f::AffineTransform{(:μ, :λ)})(x::AbstractArray) = supportdim(params(f)) == () ? f.λ .\ x .+ f.μ : f.λ \ x + f.μ
```
cscherrer, line 69: Can you help me understand these lines? This seems like implicit broadcasting. Do we still need these?

cscherrer (commented):
Hi @jeremiahpslewis, thanks again for bringing up this issue. I added a couple of comments. If we can avoid implicit broadcasting, it might only the tests that we use. But we hadn't done anything concrete to this point about broadcasting, so getting this started is still a big help.

The MeasureBase PR should be very quick, then we can make sure the tests you added here will pass.

cscherrer:
With MeasureBase 0.12.3 and the currently registered MeasureTheory, this now works:
```julia
julia> cdf.(MeasureTheory.Normal(37.5, 7.5), [10, 25])
2-element Vector{Float64}:
 0.0001228663899651518
 0.0477903522728147
```

jeremiahpslewis:
Awesome. Dropped everything but the tests.

cscherrer (approved):

cscherrer:
Looks great! Thanks again @jeremiahpslewis for spurring this ahead

PR #233, opened 2022-08-29
Dev

cscherrer:
- Rename `Affine` to `AffinePushfwd` (cc @oschulz)
- Add `mean(::AffinePushfwd)` and `std(::AffinePushfwd)`
- Use `LebesgueMeasure()` as a base measure instead of `Lebesgue()`. The idea here is that `LebesgueMeasure`  is primitive, while `Lebesgue` is more for convenience. Typical uses of `Lebesgue` are restrictions to the unit interval or positive reals. We may (to be discussed later) also consider using this to represent Lebesgue measure on an arbitrary user-deifned space.
- Add a method `logdensity_def(p::Normal, q::Normal, x)`


oschulz (approved):

docs/src/affine.md
```diff
@@ -10,7 +10,7 @@ f(z) = σ z + μ
 
 Starting with the above definition, we'll use ``z`` to represent an "un-transformed" variable, typically coming from a measure which has neither a location nor a scale parameter, for example `Normal()`.
 
-Affine transformations are often ambiguously referred as "linear transformations". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation) in the stricter algebraic sense: For a function `f` to be linear requires 
+AffinePushfwd transformations are often ambiguously referred as "linear transformations". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/AffinePushfwd_transformation#Representation) in the stricter algebraic sense: For a function `f` to be linear requires
```
cscherrer, line 13:
```suggestion
Affine transformations are often ambiguously referred as "linear transformations". In fact, an affine transformation is ["the composition of two functions: a translation and a linear map"](https://en.wikipedia.org/wiki/Affine_transformation#Representation) in the stricter algebraic sense: For a function `f` to be linear requires 
```

docs/src/affine.md
```diff
@@ -1,4 +1,4 @@
-# Affine Transformations
+# AffinePushfwd Transformations
```
cscherrer, line 1:
```suggestion
# Affine Transformations
```

PR #234, opened 2022-08-30
add beta binomial distribution

nignatiadis:
A short PR that implements the BetaBinomial distribution. This is needed for https://github.com/cscherrer/MeasureTheory.jl/pull/226, but the PR should be self-contained.

I tried to follow the code for `Binomial` but am not sure about some things (especially `rand` having a separate signature with`Tuple{I,A}}) where {I<:Integer,A})`.


src/parameterized/betabinomial.jl
```diff
@@ -0,0 +1,44 @@
...
+basemeasure(d::BetaBinomial) = CountingMeasure()
+
+testvalue(::BetaBinomial) = 0
+
+@kwstruct BetaBinomial(n, α, β)
+
+function Base.rand(rng::AbstractRNG, ::Type, d::BetaBinomial{(:n, :α, :β)})
+    rand(rng, Dists.BetaBinomial(d.n, d.α, d.β))
+end
+
```
cscherrer, line 18:
```suggestion
```

src/parameterized/betabinomial.jl
```diff
@@ -0,0 +1,44 @@
...
+function Base.rand(
+    rng::AbstractRNG,
+    ::Type,
+    d::BetaBinomial{(:n, :α, :β),Tuple{I,A}},
+) where {I<:Integer,A}
+    rand(rng, Dists.BetaBinomial(d.n, d.α, d.β))
+end
```
cscherrer, line 25:
```suggestion
function Base.rand(
    rng::AbstractRNG,
    ::Type{T},
    d::BetaBinomial{(:n, :α, :β)},
) where {T}
    k = rand(rng, T, Beta(d.α, d.β)
    return rand(rng, T, Binomial(d.n, k)
end
```

cscherrer (commented):
Thanks @nignatiadis ! I pushed a commit running JuliaFormatter, and added a suggestion for `rand`. Unless there's an efficiency benefit to going through Distributions, this seems pretty clean to me. Let me know what you think :)

nignatiadis:
Thank you @cscherrer. Definitely makes sense to code it directly (and in fact, that's exactly the implementation in Distributions.jl).

cscherrer:
Is there anything using `proxy` for this case? If nothing calls it let's leave it out until we need it. Then we need to bump the version and should be good to go

nignatiadis:
Just removed `proxy` and bumped the version!

cscherrer (approved):
Looks good! Thanks again @nignatiadis

Issue #239, opened 2022-09-13
Goals for 0.19.0


cscherrer:
- [ ] Implement transport API for most finite measures
- [ ] Factor out `ExponentialFamily` and `Tweedie` into https://github.com/cscherrer/MeasureFamilies.jl

PR #240, opened 2022-09-13
0.19

cscherrer:
Fixes https://github.com/cscherrer/MeasureTheory.jl/issues/239

This keeps TransformVariables for now. Let's wait at least until the transport API is fully functional before dropping the dependency. When we *do* drop it, we can migrate that functionality to a new package using the transport API.


oschulz (approved):
Whew, this is a big one. I think it looks Ok ?! :-) I'm not an expert on the depth of MeasureTheory.jl yet.

src/MeasureTheory.jl
```diff
@@ -8,12 +8,12 @@ using MLStyle
...
 
-using DistributionMeasures
+# using DistributionMeasures
 using TransformVariables: asℝ₊, as𝕀, asℝ, transform
 
 import Base
-import Distributions
-const Dists = Distributions
+# import Distributions
+# const Dists = Distributions
```
cscherrer, line 16: This makes it looks like we don't depend on Distributions.jl. But we do, it's just that this code is now in the file `distributions.jl`
```suggestion
```

Issue #248, opened 2023-02-11
Error when calling `rand` on a product measure


slwu89:
Hi, I encountered a strange problem when sampling from a product measure. My actual error occurred in the context of making a Markov chain, specifically from this line https://github.com/cscherrer/MeasureTheory.jl/blob/50d90d7aa42f60324a7fb9ba9dc45d4e2cac9d8e/src/combinators/chain.jl#L44

My MWE is below, using the same `mu` product measure object that was created in my erroring code, and using the same `rng` with the Xoshiro prng generator as was in the code (I suppose it is the default argument when using the `rand` method on the `Chain` object). Interestingly when calling `rand` on the product measure object without specifying the `rng` it works fine. The error is a method error from deep within the random number generation. 

```julia
using Random
using MeasureTheory
rng = ResettableRNG(Random.Xoshiro(), 6542022242862247233)
mu = ProductMeasure([Binomial(n = 990, p = 0.00995017), Binomial(n = 10, p = 0.0246901)])
rand(rng, mu)
rand(mu)
```

The error is:
```julia
ERROR: MethodError: no method matching rng_native_52(::ResettableRNG{Xoshiro, UInt64})
```

Here's my Julia info, and I'm on `MeasureTheory v0.18.1`. Thanks!

```
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.5.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 
```

slwu89:
I noticed this issue https://github.com/cscherrer/MeasureBase.jl/issues/34 but the `Binomial` objects in the product measure here are definitely the ones from MeasureTheory.jl, rather than Distributions.jl

slwu89:
Looking through the trace, this issue can be "fixed" by changing the definition here https://github.com/cscherrer/MeasureTheory.jl/blob/master/src/parameterized/binomial.jl#L44 to below (passing `rng.rng` as the first arg to `rand`).

```julia
function Base.rand(
    rng::AbstractRNG,
    ::Type,
    d::Binomial{(:n, :p),Tuple{I,A}},
[...]

PR #249, opened 2023-02-14
add method for Distributions.rand(::ResettableRNG, ::Binomial)

slwu89:
This should address #248, which actually didn't have anything to do with product measure, but rather a method for the Binomial distribution. Also added a test to check for this behavior; it would fail previously.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MeasureTheory"
 uuid = "eadaa1a4-d27c-401d-8699-e962e1bbc33b"
 authors = ["Chad Scherrer <chad.scherrer@gmail.com> and contributors"]
-version = "0.18.1"
+version = "0.18.2"
 
 [deps]
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
--- a/src/parameterized/binomial.jl
+++ b/src/parameterized/binomial.jl
@@ -23,15 +23,11 @@ end
     x ∈ (0, 1)
 end
 
-function Base.rand(rng::AbstractRNG, ::Type, d::Binomial{(:n, :p)})
-    rand(rng, Dists.Binomial(d.n, d.p))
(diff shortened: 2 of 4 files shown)


cscherrer:
Hi @slwu89 , looks like I have some catching up to do:
1. Welcome!
2. Sorry I had missed your #248 
3. Thank you for the PR!!

I like the simplicity of your approach, but it specializes for `Binomial`, which shouldn't need to be the case. In this code
```julia
function Dists.rand(rng::ResettableRNG, d::Dists.Binomial)
    rand(rng.rng, d)
end
```

we'd usually instead write
```julia
function Dists.rand(rng::ResettableRNG, ::Type{T}, d::Dists.Binomial) where {T}
    rand(rng.rng, T, d)
end
```

The idea is that the `T` is propagated down to the low-level code, where it ends up calling `rand(T)`, `randexp(T)`, or `randn(T)`. This works well in most cases, but for `ResettableRNG`s it's currently broken. I think at some point it's dispatching incorrectly. I'll have a look over the weekend; I'm guessing it can be a quick fix.

cscherrer:
Hi @slwu89 , I think I got the dispatch problems fixed. Here's your MWE now (using Julia 1.9, because that's just what I have loaded)
```julia
julia> using Random

julia> using MeasureTheory

julia> rng = ResettableRNG(Random.Xoshiro(), 6542022242862247233)
ResettableRNG(::Xoshiro, 6542022242862247233)

julia> mu = ProductMeasure([Binomial(n = 990, p = 0.00995017), Binomial(n = 10, p = 0.0246901)])
ProductMeasure([Binomial(n = 990, p = 0.00995017), Binomial(n = 10, p = 0.0246901)])

julia> rand(rng, mu)
2-element Vector{Int64}:
 12
  1
```

cscherrer:
I think this is ready to go once tests pass. Thanks for catching this and starting on the fix! I changed the solution, but it was very helpful to have the test you added to make sure things are working properly.

slwu89:
Thanks @cscherrer! Great to see the "official" solution too, helps me understand better the design of the package.

PR #250, opened 2023-02-23
Add collection of mean, var and std for distributions

theogf:
This adds direct accessors to the statistics of some given Measures

Let me know if this would be the wrong way to do this?


theogf:
That's a really weird bug: 
https://github.com/JuliaMath/MeasureTheory.jl/actions/runs/7678848125/job/20929123594?pr=250#step=347:347

Although constructed with one value, the output is different

theogf:
Looks like the issue is unrelated?

PR #260, opened 2023-08-21
MeasureBase 0.14 + other updates

cscherrer:
This is a quick fix to get MeasureTheory working with the latest MeasureBase (currently v0.14). Note that some tests have been removed; these will be reworked and added to MeasureBase.jl.

Addresses https://github.com/JuliaMath/MeasureTheory.jl/issues/259


cscherrer:
@gaurav-arya Since you brought this up, I've invited you to the repo in case you'd like to review. @mschauer I've also requested a review if you're available in case you have thoughts on this. It's a small PR so hopefully quick to review :)

src/distributions.jl
```diff
@@ -39,4 +39,3 @@ function as(d::Dists.Product, _data::NamedTuple = NamedTuple())
     as(Vector, as(v[1]), n)
 end
 
-as(m::DistributionMeasures.DistributionMeasure) = as(m.d)
```
gaurav-arya, line 42: Could you also explain this change?

src/MeasureTheory.jl
```diff
@@ -119,6 +118,9 @@ xlog1py(x, y) = x * log(1 + y)
 
 as(args...; kwargs...) = TV.as(args...; kwargs...)
```
gaurav-arya, line 119: so this `as` shadows the original `TV.as`, i.e. overloads of `as` will not become overloads of `TV.as`, is that intended?

gaurav-arya (approved):
Looks good frrom a rough scan, just left a few questions

src/transforms/corrcholesky.jl
```diff
@@ -30,9 +30,9 @@ function TV.transform_with(flag::TV.LogJacFlag, t::CorrCholesky, x::AbstractVect
     return Cholesky(U, 'U', 0), ℓ, index
 end
 
-TV.inverse_eltype(t::CorrCholesky, x::AbstractMatrix) = TV.extended_eltype(x)
+TV.inverse_eltype(t::CorrCholesky, x::AbstractMatrix) = eltype(x)
 
-TV.inverse_eltype(t::CorrCholesky, x::Cholesky) = TV.extended_eltype(x)
+TV.inverse_eltype(t::CorrCholesky, x::Cholesky) = eltype(x)
```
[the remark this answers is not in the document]
cscherrer: Within the source, we use `TV` as shorthand for `TransformVariables`. I think all the TV stuff can soon be moved to an extension, since the new transport code in MeasureBase can subsume this capability. 

TV used to have an `extended_eltype` function, but now that throws an error. I don't know details of how this was different from `eltype`, I was just mimicking the code style from that package.

src/distributions.jl
cscherrer: DistributionMeasures.jl adds methods to allow Distributions to be treated as measures. But this turns out to make things more complex in some weird ways. @oschulz and I had some discussions about this and arrived at the approach of focusing on subtypes of `AbstractMeasure`. If you want to do measure theory on something that's *not* `<:AbstractMeasure`, you can either map it into one or use a wrapper.

I don't remember the details of what was complicated about the traits-like approach. I think Oliver hit some weird corner cases in a recent MeasureBase refactoring.

src/MeasureTheory.jl
cscherrer: Good catch, yes this is a little weird. I shadowed it to avoid type piracy, in particular in cases where I defined `as` taking a `Distribution` argument. I think this will all soon go away, because we'll only need Distributions.jl for some sampling and logpdf code, and we'll move TransformVariables to an extension.

cscherrer:
Thanks @gaurav-arya . I've merged this branch into https://github.com/JuliaMath/MeasureTheory.jl/pull/240, which has some other changes we had planned. Once we're set here we can close this PR and merge that one instead.

